PR #6799, opened 2022-01-10
fix: [CI-3114]: Run-time input related enhancements in CI Steps

vardanbansal-harness:
##### Summary:

1. Bug fix for https://harness.atlassian.net/browse/CI-3114.

Context:

Currently, if a field that supports run-time input (the "parent") is on the whole run-time input, then only user can specify input in form of key-value pairs. Instead if a value for such key-value pair is made run-time input, then the parent field won't render on Run/Retry Pipeline form to provide user input.

Example:

This will be the view for `Environment Variables` on Run/Retry pipeline form when two keys in it have run-time input and other two have fixed values

image

image

2. Added missing validation error for lists.

image

3. UX consistency for input set view.

##### Screenshots:

- Jest: `retrigger jest`
- ESLint: `retrigger eslint`
- Build: `retrigger build`
- Rebase: `trigger rebase`
- Cypress: `trigger cypress`


vardanbansal-harness:
trigger cypress

yc147 (approved):

src/modules/70-pipeline/strings/strings.en.yaml
```diff
@@ -964,6 +964,7 @@ dashboardDeploymentsWidget:
   runningPipeline:
     singular: Currently running pipeline
     plural: Currently running pipelines
+runtimeInputNotSupported: '{{field}} values cannot be run-time input'
```
mtran7, line 967: I believe we remove the hyphen. All other texts say "runtime" in this file

mtran7 (approved):

src/modules/70-pipeline/strings/strings.en.yaml
vardanbansal-harness: I'll fix it in follow up PR.

PR #7011, opened 2022-01-25
fix: [CDP-20171]: commit id should be commit id / git tag

AshwinBhatkal-Harness:
##### Summary:

Changed label, placeholder and validation message for input to contain git tag

##### Jira Links:

https://harness.atlassian.net/browse/CDP-20171

##### Screenshots:

image

- Jest: `retrigger jest`
- ESLint: `retrigger eslint`
- Build: `retrigger build`
- Rebase: `trigger rebase`
- Cypress: `trigger cypress`


src/strings/strings.en.yaml
```diff
@@ -972,7 +972,7 @@ validation:
   keyRequired: Key is required
   valueRequired: Value is required
   branchName: 'Branch Name is required'
-  commitId: 'Commit Id is required'
+  commitId: 'Commit Id / Git Tag is required'
```
bhavya-sinha, line 975: I don't think modification to this file is allowed. Moreover snapshots should also update

yc147 (approved):

bhavya-sinha (approved):

src/strings/strings.en.yaml
AshwinBhatkal-Harness: This is the only place I found the validation error message and it is being used in pipeline and cd only. Do I create a new string for it? 

I ran the test to update snapshots. The snapshots would not update as they contain the id and not the actual string itself

kapilkumawat86 (approved):

kapilkumawat86 (approved):

PR #7032, opened 2022-01-25
fix: [CDC-16389]: Fixed the in,not in page crash issue in JiraApproval step

chiragp1212:
1. Fixed the page crash issue when we selected in/not-in operator for field values in approval or rejection criteria
- Placed the processFormData and processInitialValues properly. It was getting read twice and hence the string was being supplied to MultiSelect instead of string[], which lead to the crash.

2. Updated uicore to latest


kapilkumawat86 (approved):

src/modules/70-pipeline/components/PipelineSteps/Steps/JiraApproval/JiraApproval.tsx
```diff
@@ -161,7 +161,7 @@ export class JiraApproval extends PipelineStep<JiraApprovalData> {
           onUpdate?.(forUpdate)
         }}
         allowableTypes={allowableTypes}
-        onChange={values => onChange?.(this.processFormData(values))}
+        onChange={values => onChange?.(values)}
```
ishantmehtaharness, line 164: We can re-write it as `onChange={onChange}`

ishantmehtaharness (approved):
Looks good

satyam-harness (approved):

src/modules/70-pipeline/components/PipelineSteps/Steps/JiraApproval/JiraApproval.tsx
chiragp1212: could be undefined that's why the check

kapilkumawat86 (approved):

PR #7108, opened 2022-01-31
feat: [GTM-663]: update feature tooltips

nanaxuHarness:
##### Summary:

this is to update feature tooltips:
1) can display multi fail messages for button;
2) update the messages
3) by default dark theme

##### Jira Links:

https://harness.atlassian.net/browse/PLG-663

##### Screenshots:

Screen Shot 2022-02-01 at 2 15 28 PM
Screen Shot 2022-02-01 at 2 14 14 PM
Screen Shot 2022-02-01 at 2 13 24 PM
Screen Shot 2022-02-01 at 2 12 44 PM
Screen Shot 2022-02-01 at 2 10 17 PM

update:
changed the tooltip bgcolor and the flag as icon:
Screen Shot 2022-02-03 at 4 01 07 PM

- Jest: `retrigger jest`
- ESLint: `retrigger eslint`
- Build: `retrigger build`
- Rebase: `trigger rebase`
- Cypress: `trigger cypress`


abhinavrastogi-harness:
"Nice work!" sounds passive-aggressive for a rate-limit.. or is it just me? 🤔 
Also, why is "200 DEPLOYMENTS" in all caps? I think the guideline is to avoid all caps everywhere.

nanaxuHarness:
> "Nice work!" sounds passive-aggressive for a rate-limit.. or is it just me? 🤔 Also, why is "200 DEPLOYMENTS" in all caps? I think the guideline is to avoid all caps everywhere.

let me double check with PB and Dina of the wording. "200 DEPLOYMENTS" is because when we don't have customized description for any FeatureIdentifers, then we just use the plain FeatureIdentifier as the description, I can make it capitalized to look better.

nanaxuHarness:
> "Nice work!" sounds passive-aggressive for a rate-limit.. or is it just me? 🤔 Also, why is "200 DEPLOYMENTS" in all caps? I think the guideline is to avoid all caps everywhere.

according to the discussion: https://harness.slack.com/archives/G01B9SARW77/p1643729292896529
1) capitalize the feature identifier if there is no customized description;
2) keep the wording `Nice work...`

abhinavrastogi-harness (dismissed):

src/modules/10-common/components/FeatureWarning/FeatureTooltip.tsx
```diff
@@ -0,0 +1,189 @@
...
+    if (!isCore && count) {
+      return customFeatureDescription ? (
+        <Text padding={{ bottom: 'small', left: 'large' }} color={Color.WHITE}>
+          {customFeatureDescription}
+        </Text>
+      ) : (
+        limitDescription
+      )
+    }
+    const requiredPlansStr = requiredPlans.length > 0 ? requiredPlans.join(' or ') : 'upgrade'
```
tan-nhu, line 54: Should these `or`, `upgrade` be in strings.en.yaml?

src/modules/10-common/components/FeatureWarning/FeatureTooltip.tsx
```diff
@@ -0,0 +1,189 @@
...
+    }, new Map<ModuleType, CheckFeatureReturn[]>())
+  }
+
+  function getGroupedTooltips(): React.ReactNode {
+    const groupedDisabledFeatures = getGroupDisabledFeatures()
+
+    function getHeader(module: ModuleType): React.ReactNode | undefined {
+      let icon, id
+      switch (module) {
+        case 'CD':
```
tan-nhu, line 93: ModuleName.CD?

src/modules/10-common/components/FeatureWarning/FeatureTooltip.tsx
```diff
@@ -0,0 +1,189 @@
...
+          icon = 'cd-main'
+          id = 'common.purpose.cd.continuous'
+          break
+
+        case 'CI':
```
tan-nhu, line 98: ModuleName.CI, etc... for the rest of the case switch

tan-nhu (dismissed):

src/modules/10-common/components/FeatureWarning/FeatureTooltip.tsx
nanaxuHarness: we are using moduleType here, which is directly from `export interface FeatureRestrictionDetailsDTO {
  moduleType?: 'CD' | 'CI' | 'CV' | 'CF' | 'CE' | 'CORE' | 'PMS' | 'TEMPLATESERVICE'`, it's also an enum
Screen Shot 2022-02-02 at 12 33 42 PM
.

src/modules/10-common/components/FeatureWarning/FeatureTooltip.tsx
nanaxuHarness: same reason as above

src/modules/10-common/components/FeatureWarning/FeatureTooltip.tsx
nanaxuHarness: updated

tan-nhu (dismissed):

mtran7:
The flag looks odd without the stick being a more prominent color like white if the background is black. I think it'd be better to even remove the flag or place it before the text like the CI an CD icons before the text. 
image

Or if there is no indentation below for CI/CD in the last screenshot, I would just update the flag svg.
image

nanaxuHarness:
> The flag looks odd without the stick being a more prominent color like white if the background is black. I think it'd be better to even remove the flag or place it before the text like the CI an CD icons before the text. image
> 
> Or if there is no indentation below for CI/CD in the last screenshot, I would just update the flag svg. image

the flag is a unicode instead of svg, it differs with browser:
Screen Shot 2022-02-02 at 3 15 18 PM

src/modules/10-common/strings/strings.en.yaml
```diff
@@ -651,6 +651,13 @@ feature:
   keepGettingShipDone:
     title: Keep Getting Ship Done!
     requestUpgrade: Request Upgrade
+  levelUp:
+    description: Nice work! You’ve reached our plan limits.
+    limitMessage:
+      description: You've used {{limits}}
+      thisMonth: ' this month'
+      includedInPlan: ' included in the plan'
+    planMessage: 'You need {{plan}} to set up '
```
mtran7, line 660: I would exclude the spaces and use template literal strings to add strings when you're rendering the text. Otherwise, pass in whatever string goes after as a variable

nanaxuHarness: updated

src/modules/10-common/strings/strings.en.yaml
```diff
@@ -651,6 +651,12 @@ feature:
   keepGettingShipDone:
     title: Keep Getting Ship Done!
     requestUpgrade: Request Upgrade
+  levelUp:
+    description: Nice work! You’ve reached our plan limits.
+    limitMessage:
+      description: You've used {{limits}}
+      thisMonth: This Month
+    planMessage: You need {{plan}} to set up
```
mtran7, line 659: 1. Should this be "You'll"?
2. Should we say the word "License" after the license type?

 Ex. You'll need Team or Enterprise license to set up multiple organizations
image

nanaxuHarness: updated

mtran7:
image
"M" in Multiple should be capitalized

nanaxuHarness:
> image "M" in Multiple should be capitalized

it's defined by each team in `FeatureDescriptor.tsx`, hope product teams will modify them when they see it

nanaxuHarness:
retrigger jest

mtran7 (approved):

tan-nhu (approved):

anjanBalgovind (approved):

tan-nhu:
FM as this is a blocker for other teams, per Nana

PR #7165, opened 2022-02-03
fix: [PL-22373]: response processing for promise apis

abhinavrastogi-harness:
##### Summary:

We have two codepaths for making API calls.
1. via restful-react when we use any `useGetXyz` hook
2. via custom generated promises when we use any `getXyzPromise` function

In the first codepath, we have a `onResponse` callback, which checks for 401 status and handles refreshing the auth token if it's close to expiry.
But the second codepath was skipping this feature. This was causing two issues:

1. The user will not get logged out even if API returns 401
2. The user will get logged out on a subsequent API call made via `useGet` path if the token has expired.

Solution:
Move the response handler to a separate function which gets triggered via a callback for `useGet` codepath and a custom global event for the `getPromise` codepath.

- Jest: `retrigger jest`
- ESLint: `retrigger eslint`
- Build: `retrigger build`
- Rebase: `trigger rebase`
- Cypress: `trigger cypress`


reetika04 (approved):

chrisblakely01 (approved):

abhinavrastogi-harness:
retrigger titlecheck

vkbansal (approved):

PR #7176, opened 2022-02-03
fix: [CDC-16484]:  Previously checked fields now showing, deletion on uncheck, added pagespinner

Sagar-Rajput7:
##### Summary:
NG - JIRA Create Edit Flow- previously selected Jira fields are not checked in edit flows which can confuse the user .

**Solution -**

> - Synced the selected fields so now checked fields are visible
> - Delete on uncheck is also working now
> - added spinner
> - field rendering after Apply changes is also working.
> 

##### Jira Links:

https://harness.atlassian.net/browse/CDC-16484

##### Screenshots:

 Checked Fields - 
image

Spinner - 
image

- Jest: `retrigger jest`
- ESLint: `retrigger eslint`
- Build: `retrigger build`
- Rebase: `trigger rebase`
- Cypress: `trigger cypress`


src/modules/70-pipeline/components/PipelineSteps/Steps/JiraCreate/helper.ts
```diff
@@ -146,12 +146,14 @@ export const getSelectedFieldsToBeAddedInForm = (
...
 ): JiraFieldNGWithValue[] => {
-  const toReturn: JiraFieldNGWithValue[] = [...existingFields]
+  const toReturn: JiraFieldNGWithValue[] = []
   newFields.forEach(field => {
     const alreadyPresent = existingFields.find(existing => existing.name === field.name)
     const alreadyPresentKVField = existingKVFields.find(kv => kv.name === field.name)
     if (!alreadyPresent && !alreadyPresentKVField) {
       toReturn.push({ ...field, value: !isEmpty(field.allowedValues) ? [] : '' })
+    } else {
+      toReturn.push({ ...field, value: alreadyPresent !== undefined ? alreadyPresent?.value : '' })
```
chiragp1212, line 156: Can you explain the significance of this else block?

Sagar-Rajput7: This is for the deletion on uncheck purpose, here when we uncheck any field, then **"getSelectedFieldsToBeAddedInForm"** got triggered for the purpose of updating the new results.

Here in **If** I am pushing the newly added field with an empty value mapped to it.
Here in **else**, I am pushing the already existing field along with their previously mapped values.

chiragp1212: Okay. Instead of undefined we should probably do
else {
      toReturn.push({ ...field, value: alreadyPresent?.value : '' })
}

chetanvekariya-harness (approved):

src/modules/70-pipeline/components/PipelineSteps/Steps/JiraCreate/helper.ts
Sagar-Rajput7: --> **value:** is expecting a defined value so If I pass in this way then I also have to change the props and add ? therefor allowing undefined behavior there 

like this : 
image

if I am allowed to change that then this will work fine -
image

chiragp1212 (approved):

PR #7219, opened 2022-02-07
fix: [FFM-2190]: ensure percentage rollout can load from YAML

knagurski:
This PR pre-processes `FlagConfiguration` step form data to ensure that percentage rollout rules are respected when loaded from YAML. Previously, if the YAML for the variations was incomplete, or not in the correct order, the percentage rollout information would be inaccurate.

|Before|After|
|------|------|
|Before|After|

Refs [FFM-2190](https://harness.atlassian.net/browse/FFM-2190)

### Additions
- added `preProcessFormValues` function and associated tests

### Changes
- updated `FlagConfigurationStepWidget` component to call `preProcessFormValues` function as a `useMemo`


vardanbansal-harness (approved):
LGTM

tan-nhu (approved):

PR #7272, opened 2022-02-09
fix: [FFM-1805]: Fixed overflow issue and tidied up styles

chrisblakely01:
##### Summary:

Fixes an overflow issue on the flag details screen. Now truncates the text like we do in other places

##### Jira Links:

https://harness.atlassian.net/browse/FFM-1805

##### Screenshots:

Screenshot 2022-02-09 at 16 21 24

ID RE-ADDED
Screenshot 2022-02-10 at 12 32 43

- Jest: `retrigger jest`
- ESLint: `retrigger eslint`
- Build: `retrigger build`
- Rebase: `trigger rebase`
- Cypress: `trigger cypress`


src/modules/75-cf/components/FlagActivation/FlagActivationDetails.tsx
```diff
@@ -198,39 +197,18 @@ const FlagActivationDetails: React.FC<FlagActivationDetailsProps> = props => {
...
-          {featureFlag.archived && (
-            <Text inline color={Color.GREY_400} padding={{ left: 'xsmall' }} font={{ size: 'small' }}>
-              ({getString('cf.shared.archived')})
-            </Text>
-          )}
+        <Heading>
+          <Text lineClamp={1} color={Color.BLACK} font={{ variation: FontVariation.H4 }}>
+            {featureFlag.name}
+          </Text>
         </Heading>
```
knagurski, line 204: Instead of a `p` tag in the heading, can we just shift this to the heading itself?
```suggestion
        <Heading
          level={4} // is this the right level?
          lineClamp={1}
          color={Color.BLACK}
          font={{ variation: FontVariation.H4 }}
        >
          {featureFlag.name}
        </Heading>
```

src/modules/75-cf/components/IdentifierText/IdentifierText.tsx
```diff
@@ -20,34 +20,14 @@ export const IdentifierText: React.FC<IdentifierTextProps> = ({ identifier, styl
...
-        borderRadius: '4px',
-        fontSize: '12px',
-        lineHeight: '15px',
-        color: '#22222A',
-        border: '1px solid #D9DAE6',
-        ...style
-      }}
-      {...props}
-    >
-      {getString('idLabel')}
```
knagurski, line 38: Looks like this got lost

knagurski (dismissed):
Looks good, just missing the `Id:` prefix and there's an opportunity to remove some unneeded markup

src/modules/75-cf/components/FlagActivation/FlagActivationDetails.tsx
chrisblakely01: nope - lineClamp exist as a prop on Heading

src/modules/75-cf/components/IdentifierText/IdentifierText.tsx
chrisblakely01: hmm its not in the new figmas for the flag page, and imo it looks better without. But I'll put it back in, no doubt someone will nitpick about it being missing!

src/modules/75-cf/components/FlagActivation/FlagActivationDetails.tsx
knagurski: Doh... been meaning to contribute that to the `Heading` component (really `Heading` aught to be a `Text` component with a different tag and the level.

Anyway, the `Heading` needs a level and setting as `inline` will give you a `span` instead of a `p`.
```suggestion
        <Heading level={4}>
          <Text inline lineClamp={1} color={Color.BLACK} font={{ variation: FontVariation.H4 }}>
            {featureFlag.name}
          </Text>
        </Heading>
```

The markup in NGUI leaves something to be desired.

chrisblakely01: nice

src/modules/75-cf/components/IdentifierText/IdentifierText.tsx
knagurski: Yeah, it's meant to be in across the board anywhere we display an id (also the Figmas are pretty darn far from pixel perfect, so treat them more as a guideline than the law)

src/modules/75-cf/components/FlagActivation/FlagActivationDetails.tsx
```diff
@@ -197,8 +197,8 @@ const FlagActivationDetails: React.FC<FlagActivationDetailsProps> = props => {
       )}
 
       <Container>
-        <Heading>
-          <Text lineClamp={1} color={Color.BLACK} font={{ variation: FontVariation.H4 }}>
+        <Heading level={4}>
```
knagurski, line 200: I was just guessing at this being an `h4` worth double checking the hierarchy to make sure. From the UX standardisation stuff, the listing page's heading is a `h2`, so that might be correct. The level and the font variation don't have to match.

src/modules/75-cf/components/IdentifierText/IdentifierText.tsx
chrisblakely01: Figmas should be source of truth though, especially to avoid churnover in PRs

src/modules/75-cf/components/FlagActivation/FlagActivationDetails.tsx
chrisblakely01: I asked Ksenia and she said h4

Re. UX standardisation stuff - where is the source of truth for this? I keep hearing the term but I dont see where I'm supposed to go to get granular values

knagurski: I'm nearly 100% sure h4 is too deep. We don't have 3 layers of heading above this, so it's an invalid hierarchy. The h4 visual style is correct, but it's wrong for a11y.

knagurski: The UX consolidation slides are linked in the epic (sorry, I'm on my phone, so don't have the link on hand)

chrisblakely01: I will fix this now but we need to discuss when this consolidation thing is taken place. IMO this should not have been brought up during a PR for this bug fix, if consolidation is in flight it needs to be pointed out and listed in the acceptance criteria for a ticket and we need a process put in place to make sure that happens

The amount of churn on this PR for this 2 hour bug fix is crazy.

chrisblakely01: As a secondary note, UI and UX need absolutely get on the same page about this. If I get one answer about what something should be from UX and its constantly challenged by UI we have a problem

knagurski: UX consolidation is applicable to ALL PRs and was announced as such last year. We will no longer ship code that does not conform. Regarding the level of a heading, that is a not a UX concern, it's a document integrity and a11y concern.

chrisblakely01: @knagurski my point still stands - it was announced via slack but I see no plan, process or approach to actually implement it. Where does the UX consolidation scope begin/end for a given ticket? Why does it come into play with bug fix tickets, given I have an upcoming epic that addresses this exact area where it could have been addressed?

I agree its needed yes but this is not the way to go about it, as evident in the amount of churn we're experiencing at the moment

chrisblakely01:
retrigger jest

crystal-fong (approved):

willd0yle (approved):

PR #7304, opened 2022-02-11
feat: [FFM-2071]: Update CF services

knagurski:
This PR updates the CF services and makes corresponding changes to FF module code.

Refs [FFM-2071](https://harness.atlassian.net/browse/FFM-2071)

### Changes
- changes `account` to `accountIdentifier`
- changes `environment` to `environmentIdentifier`
- changes `project` to `projectIdentifier`
- changes `org` to `orgIdentifier`
- various other related changes


tan-nhu (approved):

src/modules/75-cf/components/AddKeyDialog/AddKeyDialog.tsx
```diff
@@ -79,7 +78,7 @@ const AddKeyDialog: React.FC<Props> = ({ disabled, primary, environment, onCreat
     createKey({
       identifier: values.name,
       name: values.name,
-      type: values.type
+      type: values.type === 'server' ? 'Server' : 'Client'
```
chrisblakely01, line 81: Wonder could we type this, passing `server` around as a string and then switching the case at the end feels like it could be tricky to follow if debugging for example.

But not one for now!

chrisblakely01 (approved):
IDENTIFIER all the things! 1 comment but non blocker - LGTM

src/modules/75-cf/components/AddKeyDialog/AddKeyDialog.tsx
knagurski: This one is dodgy, but it's how the backend works (according to the Swagger docs). I've flagged to the BE guys and I'm hoping we can settle on the case. Seems weird that they ensure it'll be lowercase in one place and will be capitalised in another.

kapilkumawat86 (approved):

vardanbansal-harness (approved):

PR #7332, opened 2022-02-14
fix: [PL-21420]: n/a status for non failure and non success status  in connector list

swarajcheguri-harness:
image

##### Screenshots:

- Jest: `retrigger jest`
- ESLint: `retrigger eslint`
- Build: `retrigger build`
- Rebase: `trigger rebase`
- Cypress: `trigger cypress`


swarajcheguri-harness:
retrigger titlecheck

src/modules/35-connectors/pages/connectors/views/connectivityStatus/ConnectivityStatus.tsx
```diff
@@ -218,21 +218,25 @@ const ConnectivityStatus: React.FC<ConnectivityStatusProps> = ({ data }) => {
     if (!(connectorStatus || errorMessage)) {
       return undefined
     }
+    let statusMsg = ''
+    switch (connectorStatus) {
```
avinashmadhwani02, line 222: Better way to write this - 

const statusMessageMap = {
  [ConnectorStatus.SUCCESS]: getString('success')
  [ConnectorStatus.FAILURE]: getString('failed')
}

const statusMessage = statusMessageMap[connectorStatus] || getString('na');

Pros -
More readable
we don't need switch, case and break.

swarajcheguri-harness: Since failed and success will be lower case thought of having na to be also in lower case for consistency purpose. Verified that writing na in small case is standard process. Happy to change to capitals if UX objects it. 
Regarding statusmessageMap suggestion, nothing wrong with the suggestion, but here we are creating an unnecessary temporary variable statusMessageMap and cognitive complexity in both cases is +1
image

avinashmadhwani02: Sure, you can check with UX once. I was using uppercase N/A in Audit trails, so thought it should be consistent. 

For Switch statement - You can move this code to a util method if you don't want to create an extra object. It will save cognitive complexity as well. (Adding more statuses in future will make this component big)

avinashmadhwani02 (approved):

src/modules/35-connectors/pages/connectors/views/connectivityStatus/ConnectivityStatus.tsx
abhinavrastogi-harness: +1 for maintaining  a map instead of switch-case

src/modules/35-connectors/pages/connectors/views/connectivityStatus/ConnectivityStatus.tsx
```diff
@@ -218,21 +218,25 @@ const ConnectivityStatus: React.FC<ConnectivityStatusProps> = ({ data }) => {
...
+      case ConnectorStatus.SUCCESS: {
+        statusMsg = getString('success').toLowerCase()
```
abhinavrastogi-harness, line 224: why are statuses in lower case? they should be in sentence case like "Success" and "Failure" imo.
Also, why are we converting to lowercase in JS? Why not change the string itself?

abhinavrastogi-harness (approved):

src/modules/35-connectors/pages/connectors/views/connectivityStatus/ConnectivityStatus.tsx
```diff
@@ -218,21 +220,17 @@ const ConnectivityStatus: React.FC<ConnectivityStatusProps> = ({ data }) => {
     if (!(connectorStatus || errorMessage)) {
       return undefined
     }
+    const statusMessageMap = {
+      [`${ConnectorStatus.SUCCESS}`]: getString('success'),
```
avinashmadhwani02, line 224: why do we need template string?

[`${ConnectorStatus.SUCCESS}`] can be [ConnectorStatus.SUCCESS]

src/modules/35-connectors/pages/connectors/views/connectivityStatus/ConnectivityStatus.tsx
```diff
@@ -218,21 +220,17 @@ const ConnectivityStatus: React.FC<ConnectivityStatusProps> = ({ data }) => {
...
+      [`${ConnectorStatus.FAILURE}`]: getString('failed')
+    }
+
+    const statusMsg = defaultTo(statusMessageMap[`${connectorStatus}`], getString('na'))
```
avinashmadhwani02, line 228: Same as above

src/modules/35-connectors/pages/connectors/views/connectivityStatus/__tests__/ConnectivityStatus.test.tsx
```diff
@@ -0,0 +1,50 @@
...
+        return {
+          data: {
+            status: 'SUCCESS'
+          }
+        }
+      })
+    }
+  })
+}))
+describe('connectivity status', () => {
```
avinashmadhwani02, line 22: Why are we testing everything with screenshots?

We should test flows with some logics, for eg. `executeStepVerify` method in this component

src/modules/35-connectors/pages/connectors/views/connectivityStatus/ConnectivityStatus.tsx
swarajcheguri-harness: it gives an error 
image

src/modules/35-connectors/pages/connectors/views/connectivityStatus/__tests__/ConnectivityStatus.test.tsx
swarajcheguri-harness: the intention of this component is to show pass or failed status in a column. And this was getting failed in coverage step. instead of the snapshot, I can test it by finding out the text in the container expecting the pass and fail text. which the snapshot was also is doing. Let me know what do you think

avinashmadhwani02 (approved):

PR #7342, opened 2022-02-15
fix: [PL-22694]: MonacoEditor doesn't work with CDN

mayankVermaHarness:
##### Summary:

- Added a separate webpack file to build monaco workers
- Changed the path at which the worker files are accessed in MonacoEditor component

##### Screenshots:

- Jest: `retrigger jest`
- ESLint: `retrigger eslint`
- Build: `retrigger build`
- Rebase: `trigger rebase`
- Cypress: `trigger cypress`


mayankVermaHarness:
retrigger titlecheck

mayankVermaHarness:
retrigger titlecheck

src/modules/10-common/components/MonacoEditor/MonacoEditor.tsx
```diff
@@ -52,13 +48,24 @@ const MonacoEditor = (props: ExtendedMonacoEditorProps, ref: ReactMonacoEditorRe
         'editor.background': '#f3f3fa'
       }
     })
+
+    const getUrlPrefix = () => {
```
vkbansal, line 52: There is already a function being used by microfrontends: `window.getApiBaseUrl`. it be reused instead of writing a new one?

vkbansal (changes requested):

configs/monacoWorkers.webpack.config.js
```diff
@@ -0,0 +1,33 @@
+/*
```
abhinavrastogi-harness, line 1: can we rename the file to `webpack.monaco.js`? does that make more or less sense?

package.json
```diff
@@ -25,10 +25,10 @@
     "clean": "rm -rf dist && rm -rf node_modules/.cache",
     "dev:cypress": "NODE_ENV=development TARGET_LOCALHOST=false CYPRESS=true BASE_URL=https://localhost:8080 NODE_OPTIONS=\"--max-old-space-size=6144\" webpack serve --progress --config configs/webpack.dev.js",
     "dev:cypressCoverage": "NODE_ENV=development TARGET_LOCALHOST=false CYPRESS=true CYPRESS_COVERAGE=true BASE_URL=https://localhost:8080 NODE_OPTIONS=\"--max-old-space-size=6144\" webpack serve --progress --config configs/webpack.dev.js",
-    "dev": "NODE_ENV=development webpack serve --progress --config configs/webpack.dev.js",
+    "dev": "NODE_ENV=development webpack serve --progress --config configs/webpack.dev.js --config configs/monacoWorkers.webpack.config.js",
     "watch": "yarn clean && NODE_ENV=development webpack --watch --progress --config configs/webpack.dev.js",
```
abhinavrastogi-harness, line 29: should we add the monaco config here also?

src/modules/10-common/components/MonacoEditor/MonacoEditor.tsx
```diff
@@ -52,13 +48,24 @@ const MonacoEditor = (props: ExtendedMonacoEditorProps, ref: ReactMonacoEditorRe
...
+      let urlPrefix = `${window.location.origin}${window.location.pathname}`
+      if (urlPrefix.charAt(urlPrefix.length - 1) !== '/') {
+        urlPrefix += '/'
+      }
```
abhinavrastogi-harness, line 56: in what cases will pathname not end with a `/`?
can we use `path.join()` instead of writing a custom `charAt` check? not sure if we ship `path` package today or not.

configs/monacoWorkers.webpack.config.js
mayankVermaHarness: yes, since all other webpack files are also named as such `webpack.<context>.js`. Done

src/modules/10-common/components/MonacoEditor/MonacoEditor.tsx
mayankVermaHarness: I saw a use-case where paths wouldn't always end with a `/`, or even a user could type that in the browser without the trailing slash. Can't rely on user, hence added this as a safety check. 
This is a relatively cheap check and we don't have `path` package in dependencies.

src/modules/10-common/components/MonacoEditor/MonacoEditor.tsx
mayankVermaHarness: I checked [window.getApiBaseUrl](https://github.com/harness/harness-core-ui/blob/master/src/index.html#L14)—it depends on [API_URL defined in ng-prod-manifests](https://github.com/wings-software/ng-prod-manifests/blob/master/nextgen-ui/values.yaml#L29). The value for `window.apiUrl` is `/gateway`.

As per my understanding, the CDN files need not go through this gateway path—hence we can't use `window.getApiBaseUrl`

package.json
mayankVermaHarness: Yes, adding. Just a general question: what's the `watch` script used for?

abhinavrastogi-harness: if you are using nginx on local, you can run `yarn watch` to only build ngui, without webpack-dev-server

src/modules/10-common/components/MonacoEditor/MonacoEditor.tsx
vkbansal: ok

src/modules/10-common/components/MonacoEditor/MonacoEditor.tsx
abhinavrastogi-harness: sure

abhinavrastogi-harness (approved):

vkbansal (approved):

PR #7363, opened 2022-02-16
fix: [PL-22728]: Moving HARNESS_ENABLE_NG_AUTH_UI to dev only config

avinashmadhwani02:


abhinavrastogi-harness (dismissed):

src/index.html
```diff
@@ -48,6 +47,7 @@
     </script>
     <% } %>
     <!-- Enabling ng-auth-ui -->
+    <!-- dev only flag -->
     <% if (HARNESS_ENABLE_NG_AUTH_UI) { %>
```
mayankVermaHarness, line 51: how does `HARNESS_ENABLE_NG_AUTH_UI` have its value hydrated?
IIRC, there should be reference of the same variable in `HTMLWebpackPlugin` in webpack file

avinashmadhwani02: There's a variable in Define Plugin, right? the same way we are using __DEV__

configs/webpack.prod.js
```diff
@@ -37,7 +37,8 @@ const config = {
   plugins: [
     new webpack.DefinePlugin({
       __DEV__: false,
-      __BUGSNAG_RELEASE_VERSION__: JSON.stringify(packageJson.version)
+      __BUGSNAG_RELEASE_VERSION__: JSON.stringify(packageJson.version),
+      HARNESS_ENABLE_NG_AUTH_UI: false
```
abhinavrastogi-harness, line 41: will this force-disable the flag, even if the env variable is true in SaaS?

abhinavrastogi-harness (changes requested):

configs/webpack.prod.js
avinashmadhwani02: No this will not, HARNESS_ENABLE_NG_AUTH_UI will be replaced in index.html once the build is generated

abhinavrastogi-harness (approved):

mayankVermaHarness (approved):

PR #7435, opened 2022-02-21
fix: [FFM-2312]: Fallback to default on variation if no env default serve variation exists

knagurski:
This PR fixes a bug whereby the flag listing page would trigger the error boundary when a flag existed but an env did not exist.

|Before|After|
|------|------|
|Before|After|

Refs [FFM-2312](https://harness.atlassian.net/browse/FFM-2312)

### Additions
- added `getDefaultVariation` utility function and associated tests

### Changes
- removed assumption of matching a variation based on the index of the env default serve variation


chrisblakely01 (approved):
Nice catch/fix 👍

crystal-fong (approved):

PR #7451, opened 2022-02-22
feat: [FFM-2228]: plan enforcement improvement

crystal-fong:
##### Summary:
- Using Plan enforcement SDK to determine the banner type and message to be displayed to user. This logic will replace FF's custom banners

- Bug fix to remove hidden tabs that appears when the banner is visible 

##### Jira Links:

https://harness.atlassian.net/browse/FFM-2228

##### Screenshots:

- Jest: `retrigger jest`
- ESLint: `retrigger eslint`
- Build: `retrigger build`
- Rebase: `trigger rebase`
- Cypress: `trigger cypress`


src/modules/75-cf/RouteDestinations.tsx
```diff
@@ -49,24 +48,107 @@ import CreateSecretFromYamlPage from '@secrets/pages/createSecretFromYaml/Create
...
 import { OnboardingPage } from './pages/onboarding/OnboardingPage'
+import UsageLimitBannerText from './components/UsageLimitBanner/UsageLimitBannerText'
 
 import { OnboardingDetailPage } from './pages/onboarding/OnboardingDetailPage'
 import CFTrialHomePage from './pages/home/CFTrialHomePage'
 import FeatureFlagsLandingPage from './pages/feature-flags/FeatureFlagsLandingPage'
 
+featureFactory.registerFeaturesByModule('cf', {
+  features: [FeatureIdentifier.MAUS],
+  renderMessage: (props, getString, additionalLicenseProps = {}) => {
```
chrisblakely01, line 76: Think we can move this logic to a custom hook - so we can test it

src/modules/75-cf/RouteDestinations.tsx
```diff
@@ -49,24 +48,107 @@ import CreateSecretFromYamlPage from '@secrets/pages/createSecretFromYaml/Create
...
+    const { isEnterpriseEdition, isFreeEdition, isTeamEdition } = additionalLicenseProps
+    const featuresMap = props.features
+    const monthlyActiveUsers = featuresMap.get(FeatureIdentifier.MAUS)
```
chrisblakely01, line 79: can you merge line `78/79`?

```suggestion
    const monthlyActiveUsers = props.features.get(FeatureIdentifier.MAUS)
```

Also this number could be `undefined` (for whatever reason) so might want to check for that. I think we cater for 0 below so thats good

crystal-fong:
retrigger titlecheck

crystal-fong:
trigger jest

src/modules/75-cf/RouteDestinations.tsx
crystal-fong: done

src/modules/75-cf/RouteDestinations.tsx
crystal-fong: moved to a func in a utils file

crystal-fong:
trigger cypress

crystal-fong:
trigger cypress

crystal-fong:
retrigger jest

crystal-fong:
trigger cypress

src/modules/75-cf/utils/__tests__/UsageLimitUtils.test.ts
```diff
@@ -0,0 +1,165 @@
...
+
+      expect(message).toEqual(expectedMessage)
+      expect(bannerType).toEqual(BannerType.INFO)
+    })
+
+    test('it should return correct message when usage is above 99%', () => {
+      const count = 24999
+
+      const expectedMessage =
+        'You have used 0.96% of Monthly Active Users (MAU) included in the free plan. After 25K MAUs, flag management will be restricted.'
```
chrisblakely01, line 58: This message is a bit misleading, as we're not expecting this

```suggestion
        'You have used 99% of Monthly Active Users (MAU) included in the free plan. After 25K MAUs, flag management will be restricted.'
```

src/modules/75-cf/utils/__tests__/UsageLimitUtils.test.ts
```diff
@@ -0,0 +1,165 @@
...
+        'You have used 96% of Monthly Active Users (MAU) included in the free plan. After 25K MAUs, flag management will be restricted.'
+
+      const getString = jest.fn().mockReturnValue(expectedMessage)
+      const { message, bannerType } = getBannerText(getString, additionalLicenseProps, count, limit)
+
+      expect(message).toEqual(expectedMessage)
+      expect(bannerType).toEqual(BannerType.INFO)
+    })
+
+    test('it should return correct message when usage is above 99%', () => {
```
chrisblakely01, line 54:
```suggestion
    test('it should return correct message when usage is 99%', () => {
```

src/modules/75-cf/utils/__tests__/UsageLimitUtils.test.ts
```diff
@@ -0,0 +1,165 @@
...
+      const expectedMessage = 'You have used 96% of your Monthly Active Users (MAU) subscription limit.'
+
+      const getString = jest.fn().mockReturnValue(expectedMessage)
+      const { message, bannerType } = getBannerText(getString, additionalLicenseProps, count, limit)
+
+      expect(message).toEqual(expectedMessage)
+      expect(bannerType).toEqual(BannerType.INFO)
+    })
+
+    test('it should return correct message when usage is above 99%', () => {
```
chrisblakely01, line 132:
```suggestion
    test('it should return correct message when usage is 99%', () => {
```

src/modules/75-cf/utils/__tests__/UsageLimitUtils.test.ts
```diff
@@ -0,0 +1,165 @@
...
+ * Copyright 2022 Harness Inc. All rights reserved.
+ * Use of this source code is governed by the PolyForm Shield 1.0.0 license
+ * that can be found in the licenses directory at the root of this repository, also available at
+ * https://polyformproject.org/wp-content/uploads/2020/06/PolyForm-Shield-1.0.0.txt.
+ */
+
+import { BannerType } from '@common/layouts/Constants'
+import { getBannerText } from '../UsageLimitUtils'
+
+describe('Usage Limit Utils', () => {
```
chrisblakely01, line 11: Nice set of tests, we just need some for when `count = 0` for FREE/team/enterprise and I think we're good

chrisblakely01 (changes requested):
Nice work, just a few more test cases to add and 1/2 minor comments and I think we're good to go

chrisblakely01 (approved):
HURRAYYYYY

chrisblakely01:
trigger cypress

willd0yle (approved):

src/modules/75-cf/utils/UsageLimitUtils.ts
```diff
@@ -0,0 +1,80 @@
...
+  const clientMauPlanLimit = Number(limit)
+
+  const clientMauUsagePercentage = Math.trunc((clientMauUsageCount / clientMauPlanLimit) * 100)
+  const clientMauPlanLimitFormatted = formatToCompactNumber(clientMauPlanLimit)
+
+  const showInfoBanner = clientMauPlanLimit > 0 && clientMauUsagePercentage >= 90 && clientMauUsagePercentage < 100
+  const showWarningBanner = clientMauPlanLimit > 0 && clientMauUsagePercentage >= 100
+
+  let message!: string
+  let bannerType!: BannerType
```
knagurski, line 35: Why the `!` here?

crystal-fong: stuck it in there because it made some ts warning go away but it seems to be okay without it so I'll remove it

src/modules/75-cf/utils/UsageLimitUtils.ts
```diff
@@ -0,0 +1,80 @@
...
+export const getBannerText = (
+  getString: UseStringsReturn['getString'],
+  additionalLicenseProps: Record<string, any>,
+  count?: number,
+  limit?: number
+): BannerText => {
+  const { isEnterpriseEdition, isFreeEdition, isTeamEdition } = additionalLicenseProps
+
+  const clientMauUsageCount = Number(count)
+  const clientMauPlanLimit = Number(limit)
```
knagurski, line 26: Since `count` and `limit` are optional, passing them through `Number` could result in `NaN` and throw off the rest of the function

chrisblakely01: it wont throw off the rest of the function as its caught by `clientMauPlanLimit > 0`. `NaN > 0` will return `false`

chrisblakely01: `count` & `limit` need to be converted to a number as (I think) the types are `any`. Can you suggest a better way to do this?

src/modules/75-cf/utils/UsageLimitUtils.ts
crystal-fong: without it creates
Screenshot 2022-03-02 at 11 17 47

src/modules/75-cf/utils/UsageLimitUtils.ts
knagurski: The values are used in a couple of calculations before that as well. Instead of letting it do those, you could return early, or detect early in the caller and just not call this function at all (thereby letting you set them as required).
```suggestion
  const clientMauUsageCount = Number(count)
  const clientMauPlanLimit = Number(limit)
  
  if (isNaN(clientMauUsageCount) || isNaN(clientMauPlanLimit)) {
    // return early
  }
```

If if's coz they are typed as `any`, but always are genuinely numbers, then a type assertion is a better option.

src/modules/75-cf/utils/UsageLimitUtils.ts
knagurski: The `!` essentially tells Typescript that you you've got this and don't worry. In other words, it opts out of the type checking here and opens the gate for type errors to sneak in. A better option in this file would be to change the `BannerText` interface to allow these to be optional and just return `{}` if none of the conditions match. In the caller file you need to figure out that those 2 are `undefined` and do what `featureFactory.registerFeaturesByModule` expects in that case.

src/modules/75-cf/utils/__tests__/UsageLimitUtils.test.ts
knagurski: Also with the current ability to omit `count` and `limit` in `getBannerText`, you could do with a test for each individually being `undefined` and for both being `undefined`

src/modules/75-cf/utils/UsageLimitUtils.ts
chrisblakely01: But you couldn't test that little bit of logic in `featureFactory.registerFeaturesByModule`. Thats why the goal is to encapsulate everything in this function. That being said can you not change BannerText to allow `undefined` for message, bannerType and update the default return at `line 76` to:

```
  return {
    message: undefined
    bannerType: undefined
  }
```

chrisblakely01: Or

```suggestion
  let message?: string
  let bannerType?: BannerType
```

crystal-fong: bannerType in `featureFactory.registerFeaturesByModule` can't be undefined, but in other `RouteDestinations` it seems like banners won't render if the string is empty so could just return `bannerType: BannerType.INFO`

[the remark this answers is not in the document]
crystal-fong: removed

src/modules/75-cf/utils/UsageLimitUtils.ts
```diff
@@ -0,0 +1,84 @@
...
+ * that can be found in the licenses directory at the root of this repository, also available at
+ * https://polyformproject.org/wp-content/uploads/2020/06/PolyForm-Shield-1.0.0.txt.
+ */
+
+import { BannerType } from '@common/layouts/Constants'
+import { formatToCompactNumber } from '@cf/utils/CFUtils'
+import type { UseStringsReturn } from 'framework/strings'
+
+interface BannerText {
+  message: string | undefined
```
knagurski, line 13: You can just mark this as optional. Does the same thing
```suggestion
  message?: string
```

crystal-fong: done

src/modules/75-cf/utils/UsageLimitUtils.ts
```diff
@@ -0,0 +1,84 @@
...
+
+  const clientMauUsageCount = Number(count)
+  const clientMauPlanLimit = Number(limit)
+
+  if (isNaN(clientMauPlanLimit) || isNaN(clientMauUsageCount)) {
+    return {
+      message: undefined,
+      bannerType: BannerType.INFO
+    }
+  }
```
knagurski, line 33: **TOTALLY OPTIONAL AND MAKES NO DIFFERENCE TO THE FUNCTIONALITY**, but you can swap the logic and simplify a wee bit.

```typescript
if (!isNaN(clientMauPlanLimit) && isNaN(clientMauUsageCount)) {
  // all the other stuff in the function
}

return {
  bannerType: BannerType.INFO
}
```

Note you don't need the `message: undefined` as without it `message` will be literally _undefined_

knagurski: ^ **THIS IS TOTALLY OPTIONAL AND JUST MY OPINION**

src/modules/75-cf/utils/__tests__/UsageLimitUtils.test.ts
```diff
@@ -0,0 +1,240 @@
...
+      const count = undefined
+      const expectedMessage = undefined
+      const getString = jest.fn().mockReturnValue(expectedMessage)
+
+      const additionalLicenseProps = {
+        isFreeEdition: true,
+        isEnterpriseEdition: false,
+        isTeamEdition: false
+      }
+      const { message, bannerType } = getBannerText(getString, additionalLicenseProps, count, limit)
```
knagurski, line 58: Instead of passing `undefined` for these two, you can just not pass them
```suggestion
      const { message, bannerType } = getBannerText(getString, additionalLicenseProps)
```

src/modules/75-cf/utils/__tests__/UsageLimitUtils.test.ts
```diff
@@ -0,0 +1,240 @@
...
+      const count = 24000
+      const expectedMessage = undefined
+      const getString = jest.fn().mockReturnValue(expectedMessage)
+
+      const additionalLicenseProps = {
+        isFreeEdition: true,
+        isEnterpriseEdition: false,
+        isTeamEdition: false
+      }
+      const { message, bannerType } = getBannerText(getString, additionalLicenseProps, count, limit)
```
knagurski, line 41: Here as well, no need to pass `undefined`, just don't define it.
```suggestion
      const { message, bannerType } = getBannerText(getString, additionalLicenseProps, count)
```

src/modules/75-cf/utils/__tests__/UsageLimitUtils.test.ts
crystal-fong: doneee

src/modules/75-cf/utils/__tests__/UsageLimitUtils.test.ts
crystal-fong: doneeee

src/modules/75-cf/utils/UsageLimitUtils.ts
crystal-fong: shimmied em in

chrisblakely01 (approved):

PR #7536, opened 2022-02-25
fix: [CDS-34221]: fix crash when creating new artifactory connector a…

willd0yle:
When creating a new artifactory connector after creating a new pipeline the UI crashes. This happens because the yaml/json is empty so fields are being checked that are undefined. 

The fix that we have added now uses optionals so that it is ok if the object is undefined. When saving the object we now use lodash set to create/update the properties so that will not error on save.
 
##### Screenshots:
Current error:
before

With fix:
after

- Jest: `retrigger jest`
- ESLint: `retrigger eslint`
- Build: `retrigger build`
- Rebase: `trigger rebase`
- Cypress: `trigger cypress`

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "nextgenui",
   "description": "Harness Inc",
-  "version": "0.288.5",
+  "version": "0.288.0",
   "author": "Harness Inc",
   "license": "Harness Inc",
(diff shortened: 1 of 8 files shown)


willd0yle:
retrigger titlecheck

YounixIO (dismissed):

PR #7541, opened 2022-02-25
feat: [PL-23162]: Disable 'add members' in User group for SCIM users

avinashmadhwani02:


src/modules/20-rbac/pages/UserGroupDetails/UserGroupDetails.tsx
```diff
@@ -64,6 +64,13 @@ const UserGroupDetails: React.FC = () => {
 
   useDocumentTitle([userGroup?.name || '', getString('common.userGroups')])
 
+  let membersBtnTooltipText
```
sunnykesh-harness, line 67: We have these 6 line repeated 2 times.

We can have a utils that return a stringId 'rbac.userDetails.linkToSSOProviderModal.btnDisabledTooltipText' / 'rbac.unableToEditSCIMMembership' when required.

sunnykesh-harness (changes requested):

meenakshi-harness (approved):

PR #7590, opened 2022-03-01
feat: [PL-23255]: Changing the Default Resource Group to include all …

reetika04:
…child scopes

##### Screenshots:

- Jest: `retrigger jest`
- ESLint: `retrigger eslint`
- Build: `retrigger build`
- Rebase: `trigger rebase`
- Cypress: `trigger cypress`


src/modules/20-rbac/utils/utils.tsx
```diff
@@ -116,7 +119,25 @@ export const getPermissionRequestFromProps = (
   }
 }
 
-export const getScopeBasedManagedResourceGroup = (
+export const getScopeBasedDefaultResourceGroup = (
+  scope: Scope,
+  getString: UseStringsReturn['getString']
```
avinashmadhwani02, line 124: Let's not pass `getString` method to a utility - it should be a pure function.

avinashmadhwani02: Please add a test case for this :)

avinashmadhwani02 (dismissed):

src/modules/20-rbac/utils/utils.tsx
reetika04: @avinashmadhwani02 in this case we need to return a string for label. I don't think we have any other option than this. Already added the test case.

reetika04:
trigger rebase

src/modules/20-rbac/modals/RoleAssignmentModal/views/RoleAssignmentForm.tsx
```diff
@@ -152,7 +152,10 @@ const RoleAssignmentForm: React.FC<RoleAssignmentFormProps> = ({ noRoleAssignmen
                     disabled={isAssignmentFieldDisabled(value)}
                     popoverClassName={css.selectPopover}
                     inputProps={{
-                      placeholder: getString('rbac.usersPage.selectResourceGroup')
+                      placeholder: getString('rbac.usersPage.selectResourceGroup'),
+                      // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+                      // @ts-ignore
```
abhinavrastogi-harness, line 157: what was the TS issue here?

src/modules/20-rbac/utils/utils.tsx
abhinavrastogi-harness: since this is a tsx file, we could return `<String id='xyz' />` and modify the return types accordingly?

src/modules/20-rbac/modals/RoleAssignmentModal/views/RoleAssignmentForm.tsx
reetika04: Explanation provided here [https://github.com/mui/material-ui/issues/20160](url)

src/modules/20-rbac/utils/utils.tsx
reetika04: This is used in the select component and the select option doesn't take strings

src/modules/20-rbac/modals/RoleAssignmentModal/views/RoleAssignmentForm.tsx
reetika04: The typescript doesn't have data-testid in input props

abhinavrastogi-harness: can we use the `byRole` suggestion instead?
```
use ByRole instead of ByTestid. We generally don't recommend data-testid. Most of the (interactions) tests can be implemented using ByRole.
```

avinashmadhwani02 (approved):

PR #7662, opened 2022-03-04
fix: [PL-23401]: disable token caching

abhinavrastogi-harness:
##### Summary:

Currently NGUI keeps a copy of the token in memory, which is read from local storage on every page navigation.If the token in local storage is refreshed between a page load and an API call, UI can potentially use the old in-memory token instead of the new one.

Solution: Don't store token in-memory, and always ready from local-storage to ensure reading the latest value

- Jest: `retrigger jest`
- ESLint: `retrigger eslint`
- Build: `retrigger build`
- Rebase: `trigger rebase`
- Cypress: `trigger cypress`


reetika04 (approved):

knagurski (approved):
Looks good, though some tests would be great for such an important file (beyond the scope of this PR)

mayankVermaHarness (approved):

chiragp1212 (approved):

PR #7723, opened 2022-03-08
[WIP] chore: update module federation config

vkbansal:
##### Screenshots:

- Jest: `retrigger jest`
- ESLint: `retrigger eslint`
- Build: `retrigger build`
- Rebase: `trigger rebase`
- Cypress: `trigger cypress`

### DIFF
--- a/configs/modulefederation.config.js
+++ b/configs/modulefederation.config.js
@@ -12,7 +12,7 @@ const { version } = require('fork-ts-checker-webpack-plugin')
 /**
  * These packages must be stricly shared with exact versions
  */
-const ExactSharedPackages = ['react-dom', 'react', 'react-router-dom', 'restful-react']
+const ExactSharedPackages = ['react-dom', 'react', 'react-router-dom', 'restful-react', '@harness/use-modal']
 
 const SharedPackages = [
   'formik',
--- a/configs/modulefederation.config.js
+++ b/configs/modulefederation.config.js
@@ -7,20 +7,19 @@
 
 const packageJSON = require('../package.json')
 const { pick, omit, mapValues } = require('lodash')
+const { version } = require('fork-ts-checker-webpack-plugin')
 
 /**
  * These packages must be stricly shared with exact versions
  */
-const ExactSharedPackages = [
+const ExactSharedPackages = ['react-dom', 'react', 'react-router-dom', 'restful-react']
+
+const SharedPackages = [
   'formik',
-  'react-dom',
-  'react',
-  'react-router-dom',
   '@harness/uicore',
   '@blueprintjs/core',
   '@blueprintjs/select',
-  '@blueprintjs/datetime',
-  'restful-react'
+  '@blueprintjs/datetime'
 ]
 
 module.exports = ({ enableGitOpsUI, enableSTO }) => {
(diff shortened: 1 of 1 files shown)


configs/modulefederation.config.js
```diff
@@ -7,20 +7,19 @@
 
 const packageJSON = require('../package.json')
 const { pick, omit, mapValues } = require('lodash')
+const { version } = require('fork-ts-checker-webpack-plugin')
```
knagurski, line 10: You're not actually using this

configs/modulefederation.config.js
```diff
@@ -52,7 +51,8 @@ module.exports = ({ enableGitOpsUI, enableSTO }) => {
       mapValues(pick(packageJSON.dependencies, ExactSharedPackages), version => ({
         singleton: true,
         requiredVersion: version
-      }))
+      })),
+      pick(packageJSON.dependencies, SharedPackages)
```
knagurski, line 55: I was thinking something more like this...
```typescript
      shared: {
        ...packageJSON.dependencies,
        ...ExactSharedPackages.reduce((deps, exactDepName) => ({
          ...deps,
          [exactDepName]: {
            singleton: true,
            requiredVersion: packageJSON.dependencies[exactDepName]
          }
        }))
      }
```

configs/modulefederation.config.js
```diff
@@ -7,20 +7,19 @@
...
 const { pick, omit, mapValues } = require('lodash')
+const { version } = require('fork-ts-checker-webpack-plugin')
 
 /**
  * These packages must be stricly shared with exact versions
  */
-const ExactSharedPackages = [
+const ExactSharedPackages = ['react-dom', 'react', 'react-router-dom', 'restful-react', '@harness/use-modal']
+
+const SharedPackages = [
```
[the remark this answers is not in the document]
scottyw: I could be missing your point here, but do you mean it doesn't seem valuable to have 2 lists or that it doesn't seem valuable to share thing explicitly vs sharing everything by default?

If the former, I think the goal is to separate what _must_ be shared (singletons) from what is shared for optimization reasons. The required shared stuff should be identical for all MFEs and unambiguously must be shared, so is correspondingly difficult to upgrade in a breaking way. The optionally shared things can be unshared for upgrade purposes i.e. we can unshare and upgrade independently of platform. Then we would reshare once our MFE and platform are both upgraded. That fixes the deployment problem and helps us understand what we can upgrade safely and what we can't.

If the latter, I think we need to be conscious in the MFE about our choice to share the platform version of a dep. We want to do that as much as possible but each dependency we inherit is a potential source of breakage that we need to manage. Explicit sharing lets us make that decision and take on that cost in a conscious way, and prevents new dependencies catching us out.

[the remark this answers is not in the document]
vkbansal: We started out with sharing all the dependencies, but it was removed as it causing some issues in OPA https://github.com/harness/harness-core-ui/commit/8f5d2258bbde0fbd2a3f79c1fa586312edec4d3f

configs/modulefederation.config.js
vkbansal: We started out with sharing all the dependencies, but it was removed as it causing some issues in OPA https://github.com/harness/harness-core-ui/commit/8f5d2258bbde0fbd2a3f79c1fa586312edec4d3f

scottyw:
Will these changes break existing MFE code? I'm thinking about deployment.

PR #7740, opened 2022-03-09
fix: [PIE-2678]: conditional tooltip added when stage execution is disabled

bhavya-sinha:
##### Screenshots:

- Jest: `retrigger jest`
- ESLint: `retrigger eslint`
- Build: `retrigger build`
- Rebase: `trigger rebase`
- Cypress: `trigger cypress`


chiragp1212 (dismissed):

chiragp1212 (dismissed):

dhruvUpadhyay07 (approved):

src/modules/70-pipeline/components/RunPipelineModal/RunPipelineForm.tsx
```diff
@@ -676,6 +676,10 @@ function RunPipelineFormBasic({
     },
     [yamlHandler?.getLatestYaml]
   )
+  const isStageExecutionDisabled = useCallback((): boolean => {
+    //stageExecutionData?.data is empty array when allowStageExecution is set to false in advanced tab
+    return Boolean(pipelineExecutionId) || isEmpty(stageExecutionData?.data)
```
ishantmehtaharness, line 681: Why are you converting `pipelineExecutionId` to `Boolean(pipelineExecutionId)`?
I assume `pipelineExecutionId || isEmpty(stageExecutionData?.data)` would give the same result

bhavya-sinha: If you see, i just replaced the old code with addition condition. Not sure why Boolean(pipelineExecutionId) was added, so didn't update that part.

ishantmehtaharness: got it
when `pipelineExecutionId` is not null, ex pipelineExecutionId is 'abc', then this function will return 'abc'
and disabled prop will be abc, but disabled might be expecting a boolean
This looks good
other way we can do this is `!!pipelineExecutionId || isEmpty(stageExecutionData?.data)`

ishantmehtaharness (dismissed):
Looks Good

src/modules/70-pipeline/components/RunPipelineModal/RunPipelineForm.tsx
```diff
@@ -676,6 +676,10 @@ function RunPipelineFormBasic({
     },
     [yamlHandler?.getLatestYaml]
   )
+  const isStageExecutionDisabled = useCallback((): boolean => {
```
samarth-sharma31, line 679: I don't think we should be using use callback here, since we are not passing this callback to the any optimised child components which might cause them to rerender. This will actually add performance overhead.

Ref --> https://aheadcreative.co.uk/articles/when-to-use-react-usecallback/#:~:text=When%20not%20to%20use%20useCallback,creates%20a%20new%20function%20object.

bhavya-sinha: Agree

ishantmehtaharness: I agree with @samarth-sharma31 here.

src/modules/70-pipeline/components/RunPipelineModal/RunPipelineForm.tsx
bhavya-sinha: Yeah, but that's the same thing

ishantmehtaharness (approved):
Looks Good Now

yc147 (changes requested):

dhruvUpadhyay07 (approved):

src/modules/70-pipeline/components/RunPipelineModal/RunPipelineForm.tsx
```diff
@@ -960,6 +965,7 @@ function RunPipelineFormBasic({
             placeholder={selectedStageData.allStagesSelected ? getString('pipeline.allStages') : getString('stages')}
             className={cx({ [css.stagesDropdown]: isGitSyncEnabled })}
           />
+          <HarnessDocTooltip tooltipId={stageExecutionDisabledTooltip} useStandAlone={true} />
```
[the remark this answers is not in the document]
bhavya-sinha: It doesnot work that way. {isStageExecutionDisabled() && <HarnessDocTooltip tooltipId={stageExecutionDisabledTooltip} useStandAlone={true} />}

I had to add this for data-tooltip-id as well

samarth-sharma31 (approved):

src/modules/70-pipeline/components/RunPipelineModal/RunPipelineForm.tsx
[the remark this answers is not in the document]
chiragp1212: @bhavya-sinha I think we should try and set meaningful IDs instead of undefined. So you can have 
const stageExecutionDisabledTooltip = isStageExecutionDisabled() ? 'stageExecutionDisabled' : 'stageExecutionNotDisabled'

this way we can even add different content based on the disabled state

yc147 (approved):

src/modules/70-pipeline/components/RunPipelineModal/RunPipelineForm.tsx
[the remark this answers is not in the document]
bhavya-sinha: But if you see the ticket, the requirement was not to show tooltip at all when stageexecution is enabled. 
And the tooltip content will be added by docs team

bhavya-sinha:
trigger cypress

src/modules/70-pipeline/components/RunPipelineModal/RunPipelineForm.tsx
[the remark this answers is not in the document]
yc147: @chiragp1212 Just curious to know why do we need to provide `tooltipid` at 2 places to show one tool tip?

chiragp1212 (approved):

PR #7784, opened 2022-03-11
feat: [PL-22487]: Add initial loader & text in index.html

mayankVermaHarness:
##### Summary:

Currently, before NGUI loads, the user sees a blank page for a significant amount of time (video attached, network slowed down in the video to view the effect). Current loader appears only after Javascript files have downloaded, parsed and executed.

With this PR, a loader & text appear almost immediately as it's added in `index.html`

##### Jira Links:

https://harness.atlassian.net/browse/PL-22487

##### Screenshots / Video:

BEFORE: 
https://user-images.githubusercontent.com/96057095/157873296-0e3dbeb4-3101-4b81-ac9d-781acb614357.mov

AFTER:

https://user-images.githubusercontent.com/96057095/158343600-99309f99-c005-42c9-be50-0369c4bd58f7.mov

- Jest: `retrigger jest`
- ESLint: `retrigger eslint`
- Build: `retrigger build`
- Rebase: `trigger rebase`
- Cypress: `trigger cypress`


mayankVermaHarness:
retrigger titlecheck

src/index.html
```diff
@@ -9,6 +9,25 @@
...
+        bottom: 0;
+        right: 0;
+
+        display: flex;
+        justify-content: center;
+        align-items: center;
+        text-align: center;
+      }
+      .pre-loader-text {
+        font: 18px/1.2858 Inter, sans-serif;
```
abhinavrastogi-harness, line 26: What's with the weird font size?

src/index.html
```diff
@@ -82,7 +101,61 @@
     </script>
   </head>
   <body>
-    <div id="react-root"></div>
+    <div id="react-root">
+      <div class="pre-loader-wrapper">
+        <div>
+          <svg xmlns="http://www.w3.org/2000/svg" width="32" height="32" viewBox="0 0 128 128">
```
abhinavrastogi-harness, line 107: I wonder if hosting this Svg on the cdn might be better or worse? It'll reduce the html size, but add a new network request. Probably worse.

abhinavrastogi-harness (approved):

src/index.html
mayankVermaHarness: This is shorthand for font-size (18px) with line-height (1.2858). The value is to match with the line-height of existing loader—so that transition is smooth.

src/index.html
mayankVermaHarness: I have removed SVG due to animation being blocked by Javascript. Instead, I created an optimized gif, uploaded to CDN and have referred it in `index.html now`

mayankVermaHarness:
trigger cypress

mayankVermaHarness:
retrigger typecheck

abhinavrastogi-harness (approved):

src/index.html
```diff
@@ -9,6 +9,25 @@
     <title>Harness</title>
     <link rel="icon" type="image/x-icon" href="/favicon.ico" />
     <link rel="icon" type="image/png" href="/favicon.png" />
+    <style>
+      .pre-loader-wrapper {
+        position: fixed;
+        top: 0;
+        left: 0;
+        bottom: 0;
```
avinashmadhwani02, line 17: We can use inset: 0 in place of top, left, bottom, right

mayankVermaHarness: Browser support is not too old for `inset`, hence letting it be.
image

PR #7845, opened 2022-03-15
fix: [PL-21868]: use utility function to find module presence in url

swarajcheguri-harness:
##### Screenshots:

- Jest: `retrigger jest`
- ESLint: `retrigger eslint`
- Build: `retrigger build`
- Rebase: `trigger rebase`
- Cypress: `trigger cypress`


src/modules/45-projects-orgs/modals/ProjectModal/useProjectModal.tsx
```diff
@@ -66,7 +62,7 @@ export const useProjectModal = ({
     if (!wizardData) {
       setProjectData(wizardData)
     }
-    if (wizardData && isHome) {
+    if (wizardData && !module) {
```
abhinavrastogi-harness, line 65: `isHome` and `!module` are not equivalent conditions.
`!module` will be true for all org and account scope pages. is that expected behaviour?

swarajcheguri-harness: yes, that was a bug that will be corrected here. Basically, this feature needs to be enabled if the module is not in the url, which is the case in home and org. This feature helps to show a list of modules the user needs to select after a project is created, and this action is only needed if the user is creating a project in a nonmodule space, like home or org.

abhinavrastogi-harness (approved):

PR #8007, opened 2022-03-25
feat: [PL-22654]: tooltip id for overview page

swarajcheguri-harness:
##### Screenshots:

- Jest: `retrigger jest`
- ESLint: `retrigger eslint`
- Build: `retrigger build`
- Rebase: `trigger rebase`
- Cypress: `trigger cypress`


src/modules/10-common/components/Title/TitleWithToolTipId.tsx
```diff
@@ -0,0 +1,22 @@
...
+interface ToolTipTitleProps {
+  title: string
+  toolTipId?: string
+  textProps?: TextProps
+}
+export const TitleWithToolTipId: React.FC<ToolTipTitleProps> = ({ title, toolTipId, textProps = {} }) => {
+  return (
+    <>
+      {toolTipId ? (
+        <Text font={{ variation: FontVariation.H4 }} tooltipProps={{ dataTooltipId: `${toolTipId}` }} {...textProps}>
```
abhinavrastogi-harness, line 12: doesn't `tooltipProps` already handle empty tooltip id?
why do we need a null check for tooltipid?

src/modules/10-common/components/Title/ScopedTitle.tsx
```diff
@@ -68,7 +70,7 @@ const ScopedTitle: React.FC<TitleProps> = ({ title, overrideScope }) => {
           &nbsp;
         </Text>
       )}
-      <Text font={{ variation: FontVariation.H4 }}>{`${titleInfo.label} ${titleInfo.title}`}</Text>
+      <TitleWithToolTipId title={`${titleInfo.label} ${titleInfo.title}`} toolTipId={toolTipId} />
```
abhinavrastogi-harness, line 73: can we auto-generate the tooltip id from the title?

swarajcheguri-harness: what do you mean by auto generate id from title ? title can be same for many pages, but the content might be different, Like title: ‘Overview’, Overview page of different modules will different tootltip content, and at the same time, Title: Secrets will have same tooltip content no matter in whichever module it is in. How should we go ahead with this kind of situations

src/modules/10-common/components/Title/TitleWithToolTipId.tsx
swarajcheguri-harness: it does not take care of empty tooltip id 
image

abhinavrastogi-harness (approved):

vkbansal (approved):

src/modules/10-common/components/Title/TitleWithToolTipId.tsx
chiragp1212: hey @swarajcheguri-harness I think null check is present in the tooltip framework. This might be the case where we're passing undefined as a string "undefined"
Even the HTML doesn't attach the attribute if it is undefined/null. Please check

chiragp1212: Also, there is already a component `TextWithTooltip` that solves this use case. Please check whether we can use the same and not create a new one.

swarajcheguri-harness: done

jenil-harness (approved):

vardanbansal-harness (approved):

abhinavrastogi-harness (approved):

PR #8074, opened 2022-03-30
fix: [PIE-2732]: variable screen reload and error message css

AshwinBhatkal-Harness:
##### Summary:

Fixed variable screen user abort error by increasing debounce duration and fixed error message css

##### Jira Links:

https://harness.atlassian.net/browse/PIE-2732

##### Screenshots:

Before: 

Screenshot 2022-03-30 at 10 04 44 AM

After: 

Screenshot 2022-03-30 at 10 04 17 AM

Before:

https://user-images.githubusercontent.com/96711906/160756440-e5deaf25-5a2e-45f4-b36f-a625a2a77bdb.mov

After:

https://user-images.githubusercontent.com/96711906/160756444-21b758cc-6daf-4f18-a93c-5b11b54c8a4b.mov

- Jest: `retrigger jest`
- ESLint: `retrigger eslint`
- Build: `retrigger build`
- Rebase: `trigger rebase`
- Cypress: `trigger cypress`


src/modules/70-pipeline/components/PipelineVariablesContext/PipelineVariablesContext.tsx
```diff
@@ -120,7 +120,7 @@ export function PipelineVariablesContextProvider(
       }
     },
     queryParams: { accountIdentifier: accountId, orgIdentifier, projectIdentifier },
-    debounce: 800
+    debounce: 1300
```
vkbansal, line 123: This is just delaying the the call. The actual issue is not fixed.

vkbansal: We can close the ticket for now and do a proper fix if this issue still persists.

vkbansal (approved):

src/modules/70-pipeline/components/PipelineVariablesContext/PipelineVariablesContext.tsx
AshwinBhatkal-Harness: In the current approach whenever there is any change in the variables data, we call the applyTemplate call. This gives us the resolved pipeline with which we make a call to update the variables. 

The debounce was set to 800 to delay the call and only call once the pipeline is resolved, but race conditions occurred frequently in this case. The delay to 1300 reduces the occurrence of race conditions but does not solve it completely.

If not for this, we may have to look at a different approach to call the variables api. We can call it at the close of the pane, but I assume this is required to load variable data even while interacting with step outputs

ishantmehtaharness (approved):
Looks Good

kapilkumawat86 (approved):

PR #8116, opened 2022-04-01
fix: [PIE-2738]: Use custom formik state variable to track submit count

depanshu-dhiman:
##### Summary:
The problem was that after we submit the form and see the errors, if we resolve any one form error, all the other errors vanish as well. That should not be the case. This was happening because of a formik form prop `enableReinitialize`.
From the docs:

> [enableReinitialize](https://formik.org/docs/api/formik#enablereinitialize-boolean)?: boolean
Default is false. Control whether Formik should reset the form if initialValues changes (using deep equality).

In short, all this does is resets the form on input change. Resetting the form means, our `submitCount` also becomes `0` and therefore, our condition to show errors fail since we show errors if `submitCount` is greater than `0`.
There were 2 solutions:
1. Remove this prop - which we cannot because this is crucial for applying input sets
2. Custom solution

##### Jira Links:

https://harness.atlassian.net/browse/PIE-2738

##### Screenshots:

Screenshot 2022-04-01 at 11 38 21 AM
Screenshot 2022-04-01 at 11 38 59 AM

- Jest: `retrigger jest`
- ESLint: `retrigger eslint`
- Build: `retrigger build`
- Rebase: `trigger rebase`
- Cypress: `trigger cypress`


satyam-harness (dismissed):

src/modules/10-common/utils/formikHelpers.ts
```diff
@@ -8,7 +8,15 @@
...
-export const errorCheck = (name: string, formik?: FormikContext<any>): boolean | '' | 0 | undefined =>
-  ((get(formik?.touched, name) || (formik?.submitCount && formik?.submitCount > 0)) &&
+/* _formSubmitCount is custom state var used to track submitCount.
+ * enableReinitialize prop resets the submitCount, so error checks fail.
+ * So either remove the prop (which will cause input set issues)
+ * Or use this custom var for error checks.
+ */
+
+export const errorCheck = (name: string, formik?: FormikContext<any> | any): boolean | '' | 0 | undefined =>
+  ((get(formik?.touched, name) ||
```
chiragp1212, line 18: hey @depanshu-dhiman , can we refactor this function to make it more readable? 
1. Instead of doing formik?.xyz on every line, we can do if (formik) { ... }
2. Also, if we're using the `formSubmitCount`, should we remove the `submitCount` check at least from this function

depanshu-dhiman: Thanks for the review.
1. Made it more readable as suggested 👍 
2. We cannot remove this as this is a valid property of formik and is useful when we are not using the `enableReinitialize` prop. There are many places where we have not used the `enableReinitialize` prop and for their error checks, `submitCount` does the job perfectly. Other than that, to keep it backwards compatible, I decided to not remove it as using a custom state variable is unconventional.

depanshu-dhiman:
Closed becuase develop was reverted to a previous commit.

PR #8136, opened 2022-04-01
WIP: feat: [FFM-2058]: refactor detail page template

crystal-fong:
##### Screenshots:

- Jest: `retrigger jest`
- ESLint: `retrigger eslint`
- Build: `retrigger build`
- Rebase: `trigger rebase`
- Cypress: `trigger cypress`


src/modules/75-cf/components/DetailPageTemplate/DetailPageTemplate.tsx
```diff
@@ -29,49 +30,59 @@ export interface DetailPageTemplateProps {
...
+          </>
+        }
+        content={
+          <>
+            {menuItems && menuItems}
+            {metaData && (
+              <dl className={css.metaData}>
+                <dt>{getString('environment')}</dt>
+                <dd>{metaData.envName}</dd>
+              </dl>
```
crystal-fong, line 80: is it worth putting the <dl> into it's own component?

src/modules/75-cf/components/DetailPageTemplate/DetailPageTemplate.module.scss
```diff
@@ -0,0 +1,28 @@
...
+.metaData {
+  display: flex;
+  position: absolute;
+  top: var(--spacing-xxlarge);
+  right: var(--spacing-xxlarge);
+  margin-right: 190px;
+
+  dt {
+    font-weight: bold;
+  }
```
knagurski, line 21: You might as well move this in.
```suggestion
.metaData {
  display: flex;
  position: absolute;
  top: var(--spacing-xxlarge);
  right: var(--spacing-xxlarge);
  margin-right: 190px;

  dt {
    font-weight: bold;
    
    &::after {
      content: ': ';
    }
  }
```

src/modules/75-cf/components/DetailPageTemplate/DetailPageTemplate.module.scss
```diff
@@ -0,0 +1,28 @@
...
+  dd {
+    margin-left: var(--spacing-small);
+  }
+  dl {
+    margin: var(--spacing-small);
+  }
```
knagurski, line 27: I'm not sure this does anything as it's targeting `dl` elements within `.metaData`, which itself is a `dl`

src/modules/75-cf/components/DetailPageTemplate/DetailPageTemplate.tsx
```diff
@@ -29,49 +30,59 @@ export interface DetailPageTemplateProps {
...
-        {headerExtras}
-      </Container>
-      <Container style={{ height: `calc(100% - ${HEADER_HEIGHT}px)`, overflow: 'auto', background: '#e4ebf433' }}>
-        {children}
-      </Container>
+          </>
+        }
+        content={
+          <>
+            {menuItems && menuItems}
```
knagurski, line 75: You know this makes no sense, right? :-p

src/modules/75-cf/components/DetailPageTemplate/DetailPageTemplate.tsx
```diff
@@ -29,49 +30,59 @@ export interface DetailPageTemplateProps {
   /** Optional extra components to be added into header (context menu, edit button, etc...)  */
   /** Note: Use absolute position to style it */
   headerExtras?: React.ReactNode
+  menuItems?: React.ReactNode
```
knagurski, line 33: Instead of this being a `ReactNode`, change it to be the items from the RBAC options menu.

```suggestion
  menuItems?: RbacOptionsMenuButtonProps['items']
```

src/modules/75-cf/components/DetailPageTemplate/DetailPageTemplate.tsx
knagurski, line 75: And then here you can check if there are items and output the menu (positioned correctly).
```suggestion
            {!!menuItems.length && <RbacOptionsMenuButton items={menuItems} />}
```

src/modules/75-cf/components/DetailPageTemplate/DetailPageTemplate.tsx
knagurski: My metric is usually when things get cumbersome or have their own independent functionality. Basically you know when something really aught to be pulled out. At this point it feels fine in here, but pulling it out is fine too. 100% up to you

src/modules/75-cf/components/DetailPageTemplate/DetailPageTemplate.tsx
crystal-fong: that was friday brain crystal not me 👀

src/modules/75-cf/pages/target-group-detail/components/TargetGroupHeader.tsx
```diff
@@ -6,47 +6,35 @@
...
 
-import css from './TargetGroupHeader.module.scss'
-
 export interface TargetGroupHeaderProps {
   targetGroup: Segment
-  env: EnvironmentResponseDTO
 }
 
-const TargetGroupHeader: FC<TargetGroupHeaderProps> = ({ targetGroup, env }) => {
+const TargetGroupHeader: FC<TargetGroupHeaderProps> = ({ targetGroup }) => {
```
knagurski, line 20: You should be able to get rid of this entire component and just pass the `items` array to the template

src/modules/75-cf/pages/target-group-detail/TargetGroupDetailPage.tsx
```diff
@@ -57,7 +57,6 @@ const TargetGroupDetailPage: FC = () => {
   })
 
   const {
-    data: envData,
     loading: envLoading,
     error: envError,
     refetch: refetchEnv
```
knagurski, line 62: If we don't care about the env name and just displaying the identifier, you can drop this call entirely.

src/modules/75-cf/pages/target-group-detail/TargetGroupDetailPage.module.scss
```diff
@@ -27,3 +27,9 @@
   overflow: auto;
   display: grid;
 }
+
+.optionsMenu {
+  position: absolute;
+  top: var(--spacing-medium);
+  right: var(--spacing-medium);
+}
```
knagurski, line 35: Positioning of the menu needs to be in the template, not up to each page.

src/modules/75-cf/components/DetailPageTemplate/DetailPageTemplate.module.scss
```diff
@@ -0,0 +1,24 @@
...
+ * that can be found in the licenses directory at the root of this repository, also available at
+ * https://polyformproject.org/wp-content/uploads/2020/06/PolyForm-Shield-1.0.0.txt.
+ */
+
+.metaData {
+  display: flex;
+  position: absolute;
+  top: var(--spacing-xxlarge);
+  right: var(--spacing-xxlarge);
+  margin-right: 190px;
```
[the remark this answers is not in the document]
crystal-fong: yeah it does 
image

src/modules/75-cf/pages/target-group-detail/TargetGroupDetailPage.tsx
crystal-fong: even if `envError` is used to display an error message when the page fails to load?

crystal-fong: on my local I couldn't see the difference between name of an env vs it's identifier. But if there are differences I'll just put `envData` back in

crystal-fong: what do you mean by pulling it in?

knagurski: Doing the call to get the env on each page is VERY wasteful. Having the hook do that and expose it as `envName` or something would be much handier and less intensive. Don't worry about it for now.

src/modules/75-cf/components/DetailPageTemplate/DetailPageTemplate.tsx
crystal-fong: I think I may have misunderstood what you meant by having an array of items within `menuItem`. Did you mean each  `item` is one of these options, e.g. `Delete` / `Edit`? 
image

knagurski: You do not want each page passing elements to pop out there, you want to just allow each page to pass a list of items they want to display in a menu. If the page doesn't need this, they can either pass an empty array, or pass nothing. No page should need to worry about the position of this menu or what component will be displayed.

So, in a page, they'd pass something like this.

```typescript
<DetailPageTemplate
  /* ... */
  menuItems={[
    {
      icon: 'cross',
      text: getString('delete'),
      onClick: deleteTargetGroupDialog,
      permission: {
        resource: {
          resourceType: ResourceType.ENVIRONMENT,
          resourceIdentifier: targetGroup.environment
        },
        permission: PermissionIdentifier.DELETE_FF_TARGETGROUP
      }
    }
  ]}
>
  /* ... */
</DetailPageTemplate>
```

In the template that would translate to...

```typescript
{!!menuItems.length && (
  <RbacOptionsMenuButton
    className={css.optionsMenu}
    items={menuItems}
  />
)}

File: src/modules/75-cf/pages/target-group-detail/TargetGroupDetailPage.module.scss (hunk 8 above)
Comment (crystal-fong) [reply src/modules/75-cf/pages/target-group-detail/TargetGroupDetailPage.module.scss:35]: moved to `DetailPageTemplate` scss

**Comment by crystal-fong** [comment]
retrigger coverage

File: src/modules/75-cf/components/DetailPageTemplate/DetailPageTemplate.module.scss (hunk 2 above)
Comment (crystal-fong) [reply src/modules/75-cf/components/DetailPageTemplate/DetailPageTemplate.module.scss]: moved

File: src/modules/75-cf/components/DetailPageTemplate/DetailPageTemplate.module.scss (hunk 3 above)
Comment (crystal-fong) [reply src/modules/75-cf/components/DetailPageTemplate/DetailPageTemplate.module.scss]: removed

File: src/modules/75-cf/pages/target-group-detail/components/TargetGroupHeader.tsx (hunk 6 above)
Comment (crystal-fong) [reply src/modules/75-cf/pages/target-group-detail/components/TargetGroupHeader.tsx]: done

File: src/modules/75-cf/components/DetailPageTemplate/DetailPageTemplate.tsx (hunk 4 above)
Comment (crystal-fong) [reply src/modules/75-cf/components/DetailPageTemplate/DetailPageTemplate.tsx]: done

File: src/modules/75-cf/components/DetailPageTemplate/DetailPageTemplate.tsx [hunk 10]
```diff
@@ -29,49 +30,59 @@ export interface DetailPageTemplateProps {
...
 
 export const DetailPageTemplate: React.FC<DetailPageTemplateProps> = ({
   breadcrumbs,
   title,
   subTitle,
   identifier,
-  headerExtras,
-  children
+  children,
+  menuItems,
```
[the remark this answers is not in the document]
crystal-fong: done

src/modules/75-cf/components/DetailPageTemplate/DetailPageTemplate.tsx
```diff
@@ -28,50 +32,61 @@ export interface DetailPageTemplateProps {
...
+        }
+        content={
+          <>
+            <Container className={css.optionsMenu}>
+              {!!menuItems.length && <RbacOptionsMenuButton items={menuItems} />}
+            </Container>
+            {metaData && (
+              <dl className={css.metaData}>
+                <dt>{getString('environment')}</dt>
+                <dd>{metaData.envName}</dd>
```
knagurski, line 83: Meta data is up to the page to determine. Not every page cares about the environment and some will need to display multiple bits of meta data.
```suggestion
                {Object.entries(metaData).map(([key, val]) => (
                  <Fragment key={key}>
                    <dt>{getString(key)}</dt>
                    <dd>{val}</dd>
                  </Fragment>
                )}
```

src/modules/75-cf/components/DetailPageTemplate/DetailPageTemplate.tsx
```diff
@@ -28,50 +32,61 @@ export interface DetailPageTemplateProps {
...
+              </dl>
+            )}
+          </>
+        }
+      />
+      <Page.Body className="detailPage">{children}</Page.Body>
```
knagurski, line 89: Any reason for this class?

src/modules/75-cf/pages/segment-details/SegmentDetailPage.tsx
```diff
@@ -165,24 +174,21 @@ export const SegmentDetailPage: React.FC = () => {
...
+            resource: { resourceType: ResourceType.ENVIRONMENT, resourceIdentifier: environmentIdentifier },
+            permission: PermissionIdentifier.DELETE_FF_TARGETGROUP
+          },
+          ...planEnforcementProps
+        }
+      ]}
+      metaData={{ envName: envData?.data?.name as string }}
     >
-      <Layout.Vertical height="100%" style={{ flexGrow: 1, background: 'var(--white)' }}>
+      <Layout.Vertical height="100%" className="gitSync-container">
```
knagurski, line 191: Should this come from the `css` variable?

src/modules/75-cf/components/DetailPageTemplate/DetailPageTemplate.tsx
```diff
@@ -28,50 +32,61 @@ export interface DetailPageTemplateProps {
 
   /** Optional extra components to be added into header (context menu, edit button, etc...)  */
   /** Note: Use absolute position to style it */
-  headerExtras?: React.ReactNode
+  menuItems?: RbacOptionsMenuButtonProps['items']
+  metaData?: Record<string, string>
 }
 
 export const DetailPageTemplate: React.FC<DetailPageTemplateProps> = ({
```
knagurski, line 39: It looks like there are no tests for this file. Could we get some?

src/modules/75-cf/pages/segment-details/SegmentDetailPage.tsx
crystal-fong: oops good spot

src/modules/75-cf/components/DetailPageTemplate/DetailPageTemplate.tsx
```diff
@@ -28,50 +32,61 @@ export interface DetailPageTemplateProps {
...
-          borderBottom: '.5px solid rgb(227 224 224 / 50%)'
-        }}
-      >
-        <Layout.Vertical spacing="medium">
+      <Page.Header
+        size="large"
+        breadcrumbs={
           <NGBreadcrumbs
             customPathParams={{ module: 'cf' }}
             links={breadcrumbs.map(({ title: label, url }) => ({ label, url }))}
```
knagurski, line 57: Might also be worth simplifying this by changing the type of the `breadcrumbs` prop to use the `Breadcrumb[]` type from `10-common/components/Breadcrumbs/Breadcrumbs.tsx`. You could then drop the `map`

src/modules/75-cf/components/DetailPageTemplate/DetailPageTemplate.tsx
crystal-fong: I skipped the tests because it returned 
image
presumably from the components that uses `DetailPageTemplate`?

src/modules/75-cf/components/DetailPageTemplate/DetailPageTemplate.tsx
```diff
@@ -28,50 +28,58 @@ export interface DetailPageTemplateProps {
...
+              <Container className={css.optionsMenu}>
+                <RbacOptionsMenuButton items={menuItems} />
+              </Container>
+            )}
+            {Object.entries(metaData as Record<string, string>).map(([key, val]) => (
+              <div className={css.metaData} key={key}>
+                <dt>{getString(key as keyof StringsMap)}</dt>
+                <dd>{val}</dd>
+              </div>
+            ))}
```
knagurski, line 78: Looks like `metaData` is actually marked as optional. Probably should check it actually exists first. Also, you don't have the `dl` tag and this will actually potentially put a bunch of meta data entries right on top of each other.
```suggestion
            {!!metaDataEntries.length && (
              <dl className={css.metaData}>
                {metaDataEntries.map(([key, val]) => (
                  <div key={key}>
                    <dt>{getString(key)}</dt>
                    <dd>{val}</dd>
                  </div>
                )}
              </dl>
            )}
```

And then above,
```typescript
const metaDataEntries = useMemo<[StringKeys, string]>(
  () => Object.entries(metaData || {}),
  [metaData]
)
```

src/modules/75-cf/components/DetailPageTemplate/DetailPageTemplate.tsx
```diff
@@ -28,50 +28,58 @@ export interface DetailPageTemplateProps {
 
   /** Optional extra components to be added into header (context menu, edit button, etc...)  */
   /** Note: Use absolute position to style it */
-  headerExtras?: React.ReactNode
+  menuItems?: RbacOptionsMenuButtonProps['items']
+  metaData?: Record<string, string>
```
knagurski, line 32:
```suggestion
  metaData?: Record<StringKeys, string>
```

src/modules/75-cf/components/DetailPageTemplate/DetailPageTemplate.tsx
knagurski: Yeah, we should have some tests that prove this is working in its own right.

src/modules/75-cf/components/DetailPageTemplate/DetailPageTemplate.tsx
chrisblakely01: @knagurski IIRC Crystal had changed this code based a suggestion you made last week in this PR?

knagurski: @chrisblakely01 yep, but this does not match my suggested change and introduces a few bugs

crystal-fong: hm but why is this new suggestion adding more changes rather that just tweaking what I've changed?

chrisblakely01: @knagurski actually yes it does match. Your suggestion is here https://github.com/harness/harness-core-ui/pull/8136#discussion_r844957954 and you didn't request a `dl` in that change, nor did you suggest any null checking.

You've just completely rewritten your own suggestion

knagurski: As you will see from the suggestion, I only suggested a change to the `dt` and `dd` part, not the surrounding tag. The change in the code moves the surrounding code into the loop, which will cause a bug. I did not mention the surround missing tag before as I did not spot it.

src/modules/75-cf/components/DetailPageTemplate/DetailPageTemplate.module.scss
```diff
@@ -0,0 +1,38 @@
...
+ */
+
+.metaData {
+  display: grid;
+  position: absolute;
+  top: var(--spacing-xxxlarge);
+  right: var(--spacing-huge);
+  font-size: var(--font-size-small);
+
+  grid-template-columns: repeat(2, minmax(min-content, 250px));
```
knagurski, line 15: Whoops, just tested this with one item and it sticks too far to the right. This should do the trick...

```suggestion
  grid-template-columns: repeat(auto-fit, minmax(min-content, 250px));
  max-width: 600px;
```

Screenshot 2022-04-11 at 18 28 07

knagurski (approved):
This is gonna make things so much better on every page that uses it. Great job 🙌🏻

PR #8224, opened 2022-04-06
feat: [FFM-2243]: Add percentage rollout support

knagurski:
This PR adds Percentage Rollout support to the Target Group detail page, refactors the page to use the same flags listing component and refactors the `PercentageRollout` component to be more flexible.

|Before|After|
|------|------|
|Before|After|

Refs [FFM-2243](https://harness.atlassian.net/browse/FFM-22423)


src/modules/75-cf/pages/target-group-detail/hooks/usePercentageRolloutValidationSchema.tsx
```diff
@@ -0,0 +1,41 @@
...
+                yup.object({
+                  weight: yup.number().integer().min(0).max(100).default(0).required()
+                })
+              )
+              .test(
+                'invalidTotalError',
+                getString('cf.percentageRollout.invalidTotalError'),
+                (variations?: { weight: number }[]) =>
+                  /* istanbul ignore next */
+                  (variations || []).map(({ weight }) => weight).reduce((total, weight) => total + weight, 0) === 100
```
chrisblakely01: We should be testing this logic not ignoring as its pretty important. There probably needs to be a test that exercises the form, triggers the validation and asserts the `invalidTotalError` message appears. Didn't see any test that covers this

chrisblakely01 (changes requested):

src/modules/75-cf/pages/target-group-detail/hooks/usePercentageRolloutValidationSchema.tsx
knagurski: It's actually impossible to get into the state where it'll use `[]`, but it's also impossible to convince TS of that without a bunch of casting. If this even does get in here, then the test will still resolve as failed (as it should) since the totals will still not reach 100%.

knagurski: Triggering of the error is scattered around the place, but has 100% coverage across the various levels that contribute to the error being displayed. Might be worth an all-encompassing test at a higher level though.

chrisblakely01: ah! Ok I see, I thought this was ignoring the entire line but I see you've got a decent tests around the weights. If you default the variations to `[]` at the point you intialise the form instead of checking it here would that get rid of the extra branches here? 

I guess this is one of those instances where high test coverage isn't worth shooting for

src/modules/75-cf/components/PercentageRollout/PercentageRollout.tsx
```diff
@@ -18,25 +18,29 @@ import css from './PercentageRollout.module.scss'
...
   prefix: (fieldName: string) => string
-  targetGroups: Segment[]
   variations: Variation[]
   fieldValues?: {
     variations: { variation: string; weight: string | number }[]
   }
-  bucketByAttributes: string[]
+  targetGroups?: Segment[]
+  bucketByAttributes?: string[]
+  hideOverError?: boolean
```
chrisblakely01, line 27: Hurray!

src/modules/75-cf/hooks/useFormValues.tsx
```diff
@@ -0,0 +1,37 @@
...
+    return
+  },
+  errors: {}
+})
+
+export const FormValuesProvider: FC<FormValuesProviderProps> = ({ children, values, setField, errors = {} }) => (
+  <FormValuesContext.Provider value={{ values, setField, errors }}>{children}</FormValuesContext.Provider>
+)
+
+export default function useFormValues<FormType>(): UseFormValuesPayload<FormType> {
```
chrisblakely01, line 35: Curious as to why you needed to do this instead of just getting the stuff from Formik/its own context?

src/modules/75-cf/pages/target-group-detail/hooks/usePercentageRolloutValidationSchema.tsx
knagurski: New, more-holistic, test added 👍🏻 Good shout!

src/modules/75-cf/pages/target-group-detail/components/FlagSettingsPanel/FlagSettingsForm.tsx
```diff
@@ -44,6 +44,31 @@ export interface FlagSettingsFormProps {
...
 
+function getPercentageRolloutValues(targetGroupIdentifier: string, flag: Feature): WeightedVariation[] {
+  return (
+    flag.envProperties?.rules?.find(
+      (rule: ServingRule) =>
+        rule?.serve?.distribution &&
+        rule.clauses.some(clause => clause.op === 'segmentMatch' && clause.values.includes(targetGroupIdentifier))
+    )?.serve?.distribution?.variations || []
+  )
+}
```
chrisblakely01, line 55: dont ya just love having to do stuff like this ? 😂

chrisblakely01 (commented):
Looks good, just 1/2 Q's

src/modules/75-cf/hooks/useFormValues.tsx
knagurski: The version of formik we have doesn't come with hooks. When we update to the newer version I want to double back and yank this out.

src/modules/75-cf/pages/target-group-detail/components/FlagSettingsPanel/FlagSettingsForm.tsx
knagurski: No. No I do not

src/modules/75-cf/pages/target-group-detail/components/FlagSettingsPanel/FlagsListing/FlagsListing.tsx
```diff
@@ -55,16 +67,26 @@ const FlagsListing: FC<FlagsListingProps> = ({
...
-          <AddFlagCheckboxCell flag={row.original} fieldPrefix={`flags.${row.original.identifier}`} />
-        )
+        Cell: AddFlagCheckboxCell
+      })
+    }
+
+    if (onRowDelete) {
+      cols.push({
+        Header: '',
+        width: '5%',
```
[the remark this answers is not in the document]
knagurski: I originally had them all as %, but then opted for the variations column as px for consistency between the edit and add views. The other columns looked fine in both, so I left it. Ideally the checkbox, variation and delete should be px and details should take up the rest of the space. I'm planning of pulling this out into a common component that can be used for Targets as well in the next epic. I'll deffo take the time then to pixelise the other columns. Good eye for detail!

src/modules/75-cf/pages/target-group-detail/components/FlagSettingsPanel/FlagsListing/VariationsCell.tsx
```diff
@@ -5,37 +5,91 @@
...
+  const rowValues = useMemo<FlagSettingsVariationCellFields>(() => get(values, fieldPrefix), [values, fieldPrefix])
+
+  useEffect(() => {
+    if (rowValues?.variation === PERCENTAGE_ROLLOUT_VALUE) {
+      flag.variations.forEach(({ identifier }, index) =>
+        setField(`${fieldPrefix}.percentageRollout.variations[${index}].variation`, identifier)
+      )
+    } else {
+      flag.variations.forEach((_, index) =>
+        setField(`${fieldPrefix}.percentageRollout.variations[${index}]`, undefined)
```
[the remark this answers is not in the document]
knagurski: That is a damn fine point! I could just wipe out the entire `percentageRollout` and call it a day.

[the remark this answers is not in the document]
knagurski: Done now. I flip'n love simplifications!

src/modules/75-cf/components/PercentageRollout/PercentageRollout.tsx
knagurski: I was thinking about going a step further and yanking the current error entirely and teaching pipelines how to do the validation, but I didn't want to open that can of worms!

src/modules/75-cf/constants/index.ts
```diff
@@ -104,3 +103,5 @@ export const CFSideNavProps: SidebarContext = {
 }
 
 export const DEFAULT_FLAG_GIT_REPO_PATH = '/flags.yaml'
+
+export const PERCENTAGE_ROLLOUT_VALUE = 'Percentage Rollout'
```
tan-nhu, line 107: Why this string is not in i18n/useString?

src/modules/75-cf/components/PercentageRollout/PercentageRollout.module.scss
```diff
@@ -24,7 +27,6 @@
 }
 
 .variationRow {
-  --variationColor: hotpink;
```
crystal-fong, line 27: damn why did you take that out :(

src/modules/75-cf/constants/index.ts
knagurski: This is actually the value from the server, not what is displayed.

src/modules/75-cf/components/PercentageRollout/PercentageRollout.module.scss
knagurski: No named colours allowed :-( It was just a default value which wasn't actually used.

chrisblakely01 (approved):
Good stuff 👍  And thanks for tidying the Percentage Rollout component, saves me the work! 💯

crystal-fong (approved):

PR #8231, opened 2022-04-06
feat: [FFM-24435]: Allow ForkTsCheckerWebpackPlugin to be disabled with env var

knagurski:
This PR introduces the env variable `ENABLE_FORKTSCHECKERWEBPACK_PLUGIN`, which, when set to `false`, disables the `ForkTsCheckerWebpackPlugin` for local development.

Passing during invocation:
```bash
ENABLE_FORKTSCHECKERWEBPACK_PLUGIN=false yarn dev
```

Setting in .env file
```env
ENABLE_FORKTSCHECKERWEBPACK_PLUGIN=false
```


tan-nhu (dismissed):

configs/webpack.dev.js
```diff
@@ -169,7 +170,8 @@ if (isCypress && isCypressCoverage) {
       ]
     }
   })
-} else {
+} else if (ENABLE_FORKTSCHECKERWEBPACK_PLUGIN) {
+  console.log('ForkTsCheckerWebpackPlugin ENABLED')
```
abhinavrastogi-harness, line 174: instead of logging this separately, can we add this to the table of env vars we already print?

.env.example
```diff
@@ -1,5 +1,6 @@
-BASE_URL=https://qb.harness.io/gateway
+BASE_URL=https://qa.harness.io/gateway
 TARGET_LOCALHOST=false
 ENABLE_GITOPSUI=false
 ENABLE_GOVERNANCE=false
 ENABLE_STO=false
+ENABLE_FORKTSCHECKERWEBPACK_PLUGIN=true
```
abhinavrastogi-harness, line 6: few nits around naming:
since the default behaviour is to be enabled, can we change the name to “DISABLE_xyz”? seems to make more sense logically.
can we simplify the name to something like “DISABLE_TYPECHECK”

knagurski: I was trying to keep with the current positive naming convention, but I totally agree that it's more obvious and easier to read as `DISABLE_TYPECHECK`. Much cleaner.

configs/webpack.dev.js
knagurski: Since these are logically for different things (proxy vs dev server), I've changed it to be two tables with a title.
Screenshot 2022-04-07 at 10 44 08

.env.example
knagurski: Fixed now

abhinavrastogi-harness (approved):

PR #8440, opened 2022-04-20
feat: [FFM-1777]: UI Left Bar enhancements

Kira-Jennings:
- Removed Variations & Prerequisites card backgrounds
- Standardise ID element (Removed “ID: “ text in label)
- archive option already removed
- Standardise breadcrumbs in left bar
- Add value column to variations
- Updated left bar bg
- Tidied up display of saved prerequisites
- Flag Description text position & font style

##### Jira Links:
[FFM-1777: UI Left Bar Enhancements](https://harness.atlassian.net/browse/FFM-1777)

##### Screenshots:
image
Multivariate:
image


src/modules/75-cf/components/FlagActivation/FlagActivationDetails.module.scss
```diff
@@ -98,3 +98,44 @@
     }
   }
 }
+
+.subheading {
+  color: var(--grey-800) !important;
+  font-weight: 600 !important;
+  font-size: '16px' !important;
+  line-height: '24px' !important;
```
knagurski, line 106: Instead of using specific weights, sizes and heights, use the `font` prop of `Heading` and `Text` to set a variation. You can find variations [here](https://uicore.harness.io/?path=/story/components-text--basic).

src/modules/75-cf/components/FlagActivation/FlagActivationDetails.tsx
```diff
@@ -200,14 +214,17 @@ const FlagActivationDetails: React.FC<FlagActivationDetailsProps> = props => {
             {featureFlag.name}
           </Text>
         </Heading>
+        <IdentifierText identifier={featureFlag.identifier} allowCopy lineClamp={1} />
         {featureFlag.description && (
-          <Text margin={{ bottom: 'small' }} style={{ fontSize: '13px', lineHeight: '20px', color: '#22222A' }}>
+          <Text
+            margin={{ bottom: 'small' }}
+            color={Color.GREY_800}
+            style={{ fontSize: '13px', lineHeight: '20px', fontWeight: 400 }}
```
chrisblakely01, line 222: probs want to use font variations in here too - can check with @Kseniia-Dudina if you're not sure what font thing to use

src/modules/75-cf/components/FlagActivation/FlagActivationDetails.module.scss
```diff
@@ -98,3 +98,44 @@
...
+
+.flagDetails {
+  color: var(--grey-800) !important;
+  font-size: 13px !important;
+  line-height: 20px !important;
+}
+
+.flagVariationValue {
+  padding: var(--spacing-xsmall) var(--spacing-small) !important;
+  // padding: 3px 6px !important;
```
chrisblakely01, line 136: can remove this if unneeded

chrisblakely01 (commented):
looks good, just 1 thing around font variations

src/modules/75-cf/components/FlagActivation/FlagActivationDetails.module.scss
```diff
@@ -98,3 +98,44 @@
...
+
+.btnAddPrereq {
+  font-size: 12px !important;
+  padding: 0 !important;
+}
+
+.prereqHeaders {
+  margin-bottom: var(--spacing-xsmall);
+  border-bottom: 1px solid var(--grey-300) !important;
+  padding: var(--spacing-small) 0 !important;
```
knagurski, line 125: You should be able to do all of these with the associated props on the `Layout`

src/modules/75-cf/components/FlagActivation/FlagPrerequisites.tsx
```diff
@@ -395,58 +359,72 @@ export const FlagPrerequisites: React.FC<FlagPrerequisitesProps> = props => {
...
+          const isFlagTypeBoolean = ['true', 'false'].includes(elem.variations[0])
+          const flagVariation = elem.variations[0]
+          return (
+            <Layout.Horizontal key={i} flex={{ alignItems: 'center' }}>
+              <Text width="45%" lineClamp={1} className={css.flagDetails}>
+                <VariationTypeIcon style={{ transform: 'translateY(1px)' }} multivariate={!isFlagTypeBoolean} />
+                {elem.feature}
+              </Text>
+              <Text width="45%" className={css.flagDetails}>
+                {!isFlagTypeBoolean ? flagVariation : getString(`cf.featureFlags.${flagVariation}` as StringKeys)}
```
knagurski, line 386: String concatenation like this prevents the string tools from working. Can we pull this out into a mapping function or (since it's just true/false), use a ternary to put the strings in whole?
```suggestion
                {!isFlagTypeBoolean ? flagVariation : getString(flagVariation === 'true' ? 'cf.featureFlags.true' : 'cf.featureFlags.false')}
```

src/modules/75-cf/components/FlagActivation/FlagPrerequisites.tsx
```diff
@@ -395,58 +359,72 @@ export const FlagPrerequisites: React.FC<FlagPrerequisitesProps> = props => {
...
+          )
+        })}
+      <Layout.Horizontal flex={{ justifyContent: 'flex-start' }}>
+        <RbacButton
+          data-testid="add-prerequisites-button"
+          className={css.btnAddPrereq}
+          minimal
+          intent="primary"
+          icon="small-plus"
+          text={getString('cf.shared.prerequisites')}
```
knagurski, line 418: Might be worth asking @Kseniia-Dudina if this button needs to use the _"+ New Prerequisite"_ style as it is a creation button

src/modules/75-cf/components/IdentifierText/IdentifierText.tsx
```diff
@@ -22,9 +22,6 @@ export const IdentifierText: React.FC<IdentifierTextProps> = ({ identifier, styl
 
   return (
     <Container className={css.identifierContainer}>
-      <Text inline font={{ variation: FontVariation.SMALL }} margin={{ right: 'xsmall' }}>
-        {getString('idLabel')}
-      </Text>
```
knagurski, line 27: This needs to be here. The mockups for this page are out of date and this is needed both here and in all the other places it's used.

src/modules/75-cf/strings/strings.en.yaml
```diff
@@ -146,9 +148,9 @@ featureFlags:
   customRuleMessage: There are custom rules configured for this flag, please review the rules on detail page.
   defaultServedOn: '{{defaultVariation}} when flag is ON'
   defaultServedOff: '{{defaultVariation}} when flag is OFF'
-  createdDate: 'Created: {{date}}'
-  modifiedDate: 'Modified: {{date}}'
-  prerequisitesDesc: '(what’s required before enabling flags)'
+  createdDate: 'Created {{date}}'
+  modifiedDate: 'Modified {{date}}'
```
knagurski, line 152: Might be worth checking with @Kseniia-Dudina to see if these `:` are needed. I have a feeling it's just an old mockup.

knagurski (changes requested):
Awesome work! Just a few things to do with font variations, styling and `Text`/`Heading`.

src/modules/75-cf/components/IdentifierText/IdentifierText.tsx
Kira-Jennings: What did the ticket mean for 'standardise ID element'?

knagurski: I think the ticket description was out of date. This page was out of sync with the rest, but I see from the current version on QA that it is already standardised.

src/modules/75-cf/components/FlagActivation/FlagActivationDetails.tsx
```diff
@@ -64,9 +73,9 @@ const VariationsList: React.FC<{ featureFlag: Feature; onEditSuccess: () => void
   const typeToStringMapping = useFeatureFlagTypeToStringMapping()
 
   return (
-    <Layout.Vertical padding="large" margin={{ top: 'large' }} className={css.module}>
+    <Layout.Vertical margin={{ bottom: 'xlarge' }} id="flagVariationsContainer">
```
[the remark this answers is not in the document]
Kira-Jennings: No, force of habit

src/modules/75-cf/components/FlagActivation/FlagActivationDetails.module.scss
```diff
@@ -98,3 +98,44 @@
     }
   }
 }
+
+.subheading {
+  color: var(--grey-800) !important;
+  font-weight: 600 !important;
+  font-size: '16px' !important;
+  line-height: '24px' !important;
+  margin-bottom: var(--spacing-xsmall) !important;
```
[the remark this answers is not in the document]
Kira-Jennings: To prevent repetition I chose to put this in a class since I've used it in 3 places

src/modules/75-cf/components/FlagActivation/FlagActivationDetails.tsx
Kira-Jennings: I just updated the colour here, but I will update the rest to use variations instead of inline style

src/modules/75-cf/components/FlagActivation/FlagActivationDetails.module.scss
```diff
@@ -98,3 +98,16 @@
...
+  font-size: 12px !important;
+  padding: 0 !important;
+}
+
+.flagVariationValue {
+  padding: var(--spacing-xsmall) var(--spacing-small) !important;
+  background: var(--grey-50) !important;
+  border-radius: 4px !important;
+  color: var(--grey-400) !important;
+  font-size: 10px !important;
```
knagurski, line 112: Looks like `FontVariation.TINY` would do this

knagurski (commented):
Great work and SOOOOO dang close!

src/modules/75-cf/components/VariationWithIcon/VariationWithIcon.tsx
```diff
@@ -5,17 +5,18 @@
...
 import { CFVariationColors } from '@cf/constants'
-
 interface VariationWithIconProps {
-  variation: Variation
+  variation?: Variation
   index: number
   iconStyle?: React.CSSProperties
   textElement?: React.ReactElement | string
   textStyle?: React.CSSProperties
+  textClassName?: string
```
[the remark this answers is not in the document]
Kira-Jennings: Nope, removed

src/modules/75-cf/components/FlagActivation/FlagActivationDetails.module.scss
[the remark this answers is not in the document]
Kira-Jennings: All except border-radius need it unfortunately. This is that element I found the defects on & required a classname to get around

src/modules/75-cf/components/VariationWithIcon/VariationWithIcon.tsx
```diff
@@ -39,7 +40,7 @@ export const VariationWithIcon: React.FC<VariationWithIconProps> = ({
           ...iconStyle
         }}
       ></span>
-      <Text inline margin={{ bottom: 'xsmall' }} style={{ marginBottom: 0, ...textStyle }}>
+      <Text inline color={Color.GREY_800} font={{ variation: FontVariation.BODY }} style={{ ...textStyle }}>
```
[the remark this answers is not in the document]
Kira-Jennings: Updated this

src/modules/75-cf/components/FlagActivation/FlagPrerequisites.tsx
```diff
@@ -395,58 +359,81 @@ export const FlagPrerequisites: React.FC<FlagPrerequisitesProps> = props => {
...
+          flex={{ justifyContent: 'flex-start' }}
+          margin={{ bottom: 'xsmall' }}
+          padding="none"
+          border={{ bottom: true, color: Color.GREY_300 }}
+        >
+          <Text width="45%">{getString('flag')}</Text>
+          <Text width="45%">{getString('cf.shared.variation')}</Text>
+        </Layout.Horizontal>
+      )}
+      {Boolean(featureFlag.prerequisites?.length) &&
```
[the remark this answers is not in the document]
Kira-Jennings: I had trouble trying to nest the map from the second check inside the first check 🤔

src/modules/75-cf/components/FlagActivation/FlagActivationDetails.tsx
```diff
@@ -200,14 +214,13 @@ const FlagActivationDetails: React.FC<FlagActivationDetailsProps> = props => {
             {featureFlag.name}
           </Text>
         </Heading>
+        <IdentifierText identifier={featureFlag.identifier} allowCopy lineClamp={1} />
         {featureFlag.description && (
-          <Text margin={{ bottom: 'small' }} style={{ fontSize: '13px', lineHeight: '20px', color: '#22222A' }}>
+          <Text inline margin={{ bottom: 'small' }} color={Color.GREY_800} font={{ variation: FontVariation.SMALL }}>
```
knagurski, line 219: Any reason to use `inline` here? It'll swap it from a `<p>` to a `<span>`. It looks like you still want a block-level element.

src/modules/75-cf/components/FlagActivation/FlagPrerequisites.tsx
```diff
@@ -395,20 +368,40 @@ export const FlagPrerequisites: React.FC<FlagPrerequisitesProps> = props => {
...
+            <Text width="45%">{getString('cf.shared.variation')}</Text>
           </Layout.Horizontal>
-        )}
-        <Layout.Vertical className={css.collapseFeaturesPrerequisites}>
-          {Boolean(featureFlag.prerequisites?.length) &&
-            featureFlag.prerequisites?.map((elem, i) => (
-              <Layout.Horizontal key={i} flex padding="medium">
-                <Text>{elem.feature}</Text>
-                <Text>{elem.variations[0]}</Text>
+          {featureFlag.prerequisites?.map((elem, i) => {
```
knagurski, line 391: **No need to change**, but if you don't use `Boolean`, you might find that typescript understands that `prerequisites` exists.
```typescript
{!!featureFlag.prerequisites?.length && (
  <>
    { /* ... */ }
    {featureFlag.prerequisites.map((prereq, index) => {
      // ...
    }
  </>
)}
```

[Here's a wee snippet to illustrate](https://www.typescriptlang.org/play?#code/PTAEHUFMBsGMHsC2lQBd5oBYoCoE8AHSAZVgCcBLA1UABWgEM8BzM+AVwDsATAGiwoBnUENANQAd0gAjQRVSQAUCEmYKsTKGYUAbpGF4OY0BoadYKdJMoL+gzAzIoz3UNEiPOofEVKVqAHSKymAAmkYI7NCuqGqcANag8ABmIjQUXrFOKBJMggBcISGgoAC0oACCbvCwDKgU8JkY7p7ehCTkVDQS2E6gnPCxGcwmZqDSTgzxxWWVoASMFmgYkAAeRJTInN3ymj4d-jSCeNsMq-wuoPaOltigAKoASgAywhK7SbGQZIIz5VWCFzSeCrZagNYbChbHaxUDcCjJZLfSDbExIAgUdxkUBIursJzCFJtXydajBDIKMjJBhLABijBGAG9FCUSgQnE4AI7sITyfQAfny-XYiGk3wA2gBdRQAX2CCE4ghoqyF9IYIwAvMzWfMOZBubyFAVQOKAIy8ABMvAAzNK5YoEaAABSrALs5EGuRG-kBdycZixACUoG1rNd7q5PK9+gCyXgZAAojTME68KANQA+NGK+DuX3wZipwOB2XBFQ4ZoeMheRDx5zA9g0L5uMzMdjqyD8WDQdSJBjAvQiLwAIgTqwYiAW+mHSWxw-ADlQAHJhAA5SASYdBFQAeS+ZHegk7WkgR1QjgUrmkaactZ0wyw+hQCEQ0OEl2bEnj0REBg42IWJhWA4HgghKIA).

src/modules/75-cf/components/FlagActivation/FlagActivationDetails.tsx
```diff
@@ -45,9 +46,24 @@ interface FlagActivationDetailsProps {
...
+      flex={{ alignItems: 'center', justifyContent: 'space-between' }}
+      spacing="xsmall"
+      className={css.variationItem}
+    >
+      <Layout.Horizontal flex={{ alignItems: 'center' }} spacing="xsmall">
+        <VariationWithIcon
+          variation={variation}
+          index={index}
+          color={Color.GREY_800}
+          textStyle={{ fontSize: '13px' }}
```
knagurski, line 61: Looks like this is a non-standard font size. Is there a token we can use? I know @Kseniia-Dudina is usually dead-against non-even font sizes. Might be a case of the mockup being incorrect. Worth asking

src/modules/75-cf/components/FlagActivation/FlagPrerequisites.tsx
Kira-Jennings: Updated this to remove Boolean

src/modules/75-cf/components/FlagActivation/FlagPrerequisites.tsx
```diff
@@ -397,58 +369,55 @@ export const FlagPrerequisites: React.FC<FlagPrerequisitesProps> = props => {
...
+    <Layout.Vertical margin={{ bottom: 'medium', top: 'medium' }}>
+      <Layout.Vertical margin={{ bottom: 'medium' }}>
+        <Heading level={5} font={{ variation: FontVariation.H5 }}>
+          <StringWithTooltip stringId="cf.shared.prerequisites" tooltipId="ff_ffPrerequisites_heading" />
+        </Heading>
+        <Text font={{ variation: FontVariation.SMALL_SEMI }} color={Color.GREY_500}>
+          {getString('cf.featureFlags.prerequisitesDesc')}
+        </Text>
+      </Layout.Vertical>
+      {featureFlag.prerequisites?.length && (
```
knagurski, line 381: Little gotcha here. If `featureFlag.prerequisites.length` is `0`, it'll actually output that on screen. Use a `!!` to make sure it's a boolean at this stage
```suggestion
      {!!featureFlag.prerequisites?.length && (
```

src/modules/75-cf/components/FlagActivation/__tests__/FlagPrerequisiteItem.test.tsx
```diff
@@ -0,0 +1,58 @@
...
+      expect(screen.getByText('Test_Bool_1')).toBeInTheDocument()
+      expect(screen.getByTestId('prerequisiteMenuBtn')).toBeInTheDocument()
+    })
+
+    userEvent.click(screen.getByTestId('prerequisiteMenuBtn'))
+
+    await waitFor(() => expect(document.querySelector('#editPrerequisite')).toBeInTheDocument())
+    await waitFor(() => expect(document.querySelector('#deletePrerequisite')).toBeInTheDocument())
+
+    expect(container).toMatchSnapshot()
```
knagurski, line 56: Can we drop this snapshot test? Personally I think they're more trouble than they're worth.

src/modules/75-cf/components/FlagActivation/__tests__/FlagPrerequisiteItem.test.tsx
```diff
@@ -0,0 +1,58 @@
...
+    )
+    await waitFor(() => {
+      expect(screen.getByText('Test_Bool_1')).toBeInTheDocument()
+      expect(screen.getByTestId('prerequisiteMenuBtn')).toBeInTheDocument()
+    })
+
+    userEvent.click(screen.getByTestId('prerequisiteMenuBtn'))
+
+    await waitFor(() => expect(document.querySelector('#editPrerequisite')).toBeInTheDocument())
+    await waitFor(() => expect(document.querySelector('#deletePrerequisite')).toBeInTheDocument())
```
knagurski, line 54: Instead of using an id, try to use things that the user can actually see. In this case you should be able to get the `edit` and `delete` by the text. Even better is if you use `getByRole` to select them. I can't remember off hand what the markup is, but if they are buttons, you can use `screen.getByRole('button', { name: 'edit' })`, if they're links, swap the `'button'` for `'link'`. You can also pass a regex as the name if required.

knagurski, line 54: You can also put these both in the same `waitFor` block.

src/modules/75-cf/components/VariationWithIcon/VariationWithIcon.tsx
```diff
@@ -39,7 +43,13 @@ export const VariationWithIcon: React.FC<VariationWithIconProps> = ({
           ...iconStyle
         }}
       ></span>
-      <Text inline margin={{ bottom: 'xsmall' }} style={{ marginBottom: 0, ...textStyle }}>
+      <Text
+        inline
+        margin={{ bottom: 'xsmall' }}
+        color={color}
+        font={{ size: fontSize }}
+        style={{ marginBottom: 0, ...textStyle }} // will override fontSize if set
```
knagurski, line 51: Looks like it's setting `xsmall` `margin` with the `margin` prop and then unsetting it with an inline style.

Kira-Jennings: This shouldn't be here, I think it's an incorrect merge. However, this is how the component existed before I fixed it so I've only improved it

src/modules/75-cf/components/FlagActivation/__tests__/FlagPrerequisiteItem.test.tsx
Kira-Jennings: Like the statements above, yes

Kira-Jennings: This doesn't work due to the location the Popover is inserted into the DOM, which is nowhere near the HTML that calls it. This is the _only_ way to get it to recognise these buttons

knagurski: When a component is rendered in a test it is wrapped in a container (a `div` by default), which acts as the root of the document. When something creates a new render root (like a popover portal), it does so as a child of that container and a sibling of the component. The `screen` object acts at the container level, not at the component level, so can pick up everything, including the popover. You can have a look at the render output by destructuring the `debug` method out of the return of the `render` call; just call that method whenever you want to see what the document looks like.
```typescript
const { debug } = render(/* ... */)

// ...

debug()
```

Anyway, I had a look-e-see and it looks like you'll hit 2 issues with getting a handle on this with `getByRole`. The first is that the button is rendered as an `<a>` without a `href`, which browsers (and jsdom) treat as having no `role`. We can fix this by adding `role: 'link'` to the `RbacMenuOptionsButton` items.
```typescript
[
  {
    role: 'link',
    icon: 'edit',
    text: getString('edit'),
    onClick: () => handlePrerequisiteInteraction('edit', prerequisite),
    disabled: flagArchived,
    permission: permission
  },
  {
    role: 'link',
    icon: 'cross',
    text: getString('delete'),
    onClick: () => handlePrerequisiteInteraction('delete', prerequisite),
    disabled: flagArchived,
    permission: permission
  }
]
```

The next issue is that jsdom picks up the `desc` tag of the SVG. That means that there are then 2 nodes in the DOM that have the accessible text content of `edit`. That means `getByText` will fail as it will find both. Luckily if we add the `role` to the items, we can then use `getByRole` and use a regex to target.
```suggestion
    await waitFor(() => {
      expect(screen.getByRole('link', { name: /edit/ })).toBeInTheDocument()
      expect(screen.getByRole('link', { name: /delete/ })).toBeInTheDocument()
    })
```

Kira-Jennings: Nice

knagurski: lol, sorry for the long ramble, it was more for gathering my own thoughts on the approach rather than anything to do with this PR 🤪

src/modules/75-cf/components/FlagActivation/FlagPrerequisiteItem.tsx
```diff
@@ -19,7 +19,7 @@ interface PrerequisiteItemProps {
   prerequisite: Prerequisite
   permission?: Omit<PermissionsRequest, 'permissions'> & { permission: PermissionIdentifier }
   flagArchived: boolean
-  handlePrerequisiteInteraction: (action: string, prereq: Prerequisite) => void
+  handlePrerequisiteInteraction: (action: 'edit' | 'delete', prereq: Prerequisite) => void
```
knagurski, line 22: Nice use of narrowing! I actually prefer this over enums personally 👍🏻

knagurski (approved):
Great job sticking with this!

chrisblakely01 (dismissed):

chrisblakely01 (approved):

PR #8729, opened 2022-05-07
fix: [CDS-36473]: fixed time filter reset issue for (home, services, overview) dashboards.

Sagar-Rajput7:
##### Summary:

When the user navigates between different views (Services, Overview, Home) the system doesn’t remember the time filter that was selected (within the same session), and every time the user goes back to the same view, the filter resets.

- Added useSessionStorage hook
- stored key-value pair for timeRange on sessionStorage
- sessionStorage will clear on re-login/new-login, tab close, origin (https, port etc. ) change etc.

##### Jira Links:

https://harness.atlassian.net/browse/CDS-36473

##### Screenshots:

- Jest: `retrigger jest`
- ESLint: `retrigger eslint`
- Build: `retrigger build`
- Rebase: `trigger rebase`
- Cypress: `trigger cypress`


Sagar-Rajput7:
retrigger titlecheck

src/modules/10-common/hooks/useSessionStorage.ts
```diff
@@ -0,0 +1,75 @@
...
+ * Use if you need to store something on session storage :
+ *
+ * keyName: key to identify specific data in sessionStorage
+ *
+ * defaultValue : default value in case of empty session storage or first time input
+ *
+ * sessionStorage data will clear if - tab closed, origin changed (https, port etc.), opening in new tab from current tab etc.
+ */
+
+export const useSessionStorage = <T>(
```
abhinavrastogi-harness, line 27: we already have existing hooks for session storage.
In fact, we should consider using the Preference Framework, instead of directly writing to session storage.

abhinavrastogi-harness (changes requested):

src/modules/10-common/hooks/useSessionStorage.ts
Sagar-Rajput7: yes, we can use `useLocalStorage` hook (with window.sessionStorage as props) thanks for suggesting this.

And regarding using Preference Framework - our use case here in this ticket is to clear sessionStorage after logout but with `usePreferenceStore` we work more around storing after logout, so we can't use that here I guess.

Sagar-Rajput7:
dockerize

src/modules/10-common/hooks/useSessionStorage.ts
Sagar-Rajput7: Following this Ticket in this branch - https://github.com/harness/harness-core-ui/pull/9003

**Closing this one**

- used existing useLocalStorageHook 
- existing secureStorage helped with clearing storage after logout

PR #8754, opened 2022-05-09
fix: [FFM-1912]: Chart bar graph not showing on metrics tab

laura-mannus:
##### Summary:

When comparing the data coming back from the API that we display on the bar graph it would always return false for dates with a leading 0 i.e May 08 as the data from the API and the data we compared against was a different format. I remedied this by removing any leading zeros on the incoming metrics to match the category format. Therefore we got a fair comparison and the data was able to be displayed. 

To ensure that only leading zeros would be replaced I replace the space AND the zero, therefore any numbers with a trailing zero ie 30 will not be effected.

##### Jira Links:

(https://harness.atlassian.net/browse/FFM-1912)

##### Screenshots:

Pasted Graphic 5
INCOMING METRICS: Displayed as "May 08"

4 May 6
CATEGORIES WE WERE COMPARING AGAINST: Displaying as "May 9"

Pasted Graphic 7

When compared this way the statement would always return false for dates that began with a leading 0. Therefore not displaying their data.

Activity

However replacing any space followed by a 0 with just a space, meant that we could accurately compare dates and the data shows accurately as well. Also dates such as May 10, May 20 etc aren't effected.

Flag Evaluations (13)

- Jest: `retrigger jest`
- ESLint: `retrigger eslint`
- Build: `retrigger build`
- Rebase: `trigger rebase`
- Cypress: `trigger cypress`


laura-mannus:
I am aware that this fix is manipulating data and may not be the best approach. My other approach is to change the categories we display on the front end to reflect the same as the dates received from the API. This'll mean displaying the dates as May 08 etc on the bar graph. If this is the preferred way, please let me know and I can change the PR.

laura-mannus:
retrigger titlecheck

laura-mannus:
retrigger titlecheck

chrisblakely01 (approved):
looks like a pretty decent fix to get us out the door. Nice work 👍

PR #8755, opened 2022-05-09
feat: [STO-4464] STO Nav

jsentongo:
##### Screenshots:

- Jest: `retrigger jest`
- ESLint: `retrigger eslint`
- Build: `retrigger build`
- Rebase: `trigger rebase`
- Cypress: `trigger cypress`

### DIFF
--- a/src/modules/85-sto-steps/components/STOSideNav/STOSideNav.tsx
+++ b/src/modules/85-sto-steps/components/STOSideNav/STOSideNav.tsx
@@ -15,6 +15,7 @@ import { SidebarLink } from '@common/navigation/SideNav/SideNav'
 import { ProjectSelector } from '@projects-orgs/components/ProjectSelector/ProjectSelector'
 import { useAppStore } from 'framework/AppStore/AppStoreContext'
 import { useStrings } from 'framework/strings'
+import ProjectSetupMenu from '@common/navigation/ProjectSetupMenu/ProjectSetupMenu'
 import css from './STOSideNav.module.scss'
 
 export default function STOSideNav(): React.ReactElement {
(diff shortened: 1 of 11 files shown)


src/modules/85-sto-steps/components/STOSideNav/STOSideNav.tsx
```diff
@@ -116,6 +117,7 @@ const TabPanel: React.FC<TabPanelProps> = ({ isProjectMode }) => {
               <SidebarLink label={getString('stoSteps.targets.testTargets')} to={routes.toSTOTargets({ accountId })} />
             </>
           )}
+          <ProjectSetupMenu module={'sto'} />
```
chris-harness, line 120: Seems like this should only be shown in project mode

src/modules/85-sto-steps/strings/strings.en.yaml
```diff
@@ -1,3 +1,4 @@
+continuous: '{{$.common.purpose.sto.continuous}}'
```
chris-harness, line 1: We shouldn't be duplicating strings, which is what this is doing in an indirect way

src/stringTypes.ts
```diff
@@ -8895,6 +8897,10 @@ export interface StringsMap {
    * @deprecated migrate this string to module level file
    */
   secureToken: string
+  /**
+   * @deprecated migrate this string to module level file
+   */
+  securityText: string
```
chris-harness, line 8903: This shouldn't be put somewhere that is deprecated, and I think we already have this string elsewhere

chris-harness (changes requested):

PR #8816, opened 2022-05-11
fix: [PL-25102]: Prevent gitSync filters from appearing on Account/Org page

mayankVermaHarness:
##### Summary:

Due to preference framework, gitSyncAPI started using Org/Project from saved preferences. Due to this, gitFilters were visible on account level page as well—which is wrong.

Re-instated the gitSync API to use Org/Project from path only. Now, gitFilters are only visible at project scope (& not at account scope)

##### Jira Links:

https://harness.atlassian.net/browse/PL-25102

##### Screenshots:

BEFORE (buggy behaviour):
image

AFTER:
image

- Jest: `retrigger jest`
- ESLint: `retrigger eslint`
- Build: `retrigger build`
- Rebase: `trigger rebase`
- Cypress: `trigger cypress`


mayankVermaHarness:
retrigger titlecheck

src/framework/AppStore/AppStoreContext.tsx
```diff
@@ -190,9 +190,14 @@ export function AppStoreProvider(props: React.PropsWithChildren<unknown>): React
 
   // update gitSyncEnabled when selectedProject changes
   useEffect(() => {
-    if (projectIdentifier) {
+    // For gitSync, using path params instead of project/org from PreferenceFramework
+    if (projectIdentifierFromPath) {
```
avinashmadhwani02, line 194: @mayankVermaHarness projectIdentifier was initialised by using projectIdentifierFromPath only before. What is the difference between both?

I think the problem is `projectIdentifier` and `orgIdentifiers` are local variables, and useEffect will not be called when the value of local variable changes. 
Should we move this to state? Because I can see `projectIdentifier` and `orgIdentifiers`  are used in one more useEffect at line number 222.

mayankVermaHarness:
retrigger jest

mayankVermaHarness:
retrigger coverage

abhinavrastogi-harness (approved):

PR #8892, opened 2022-05-17
feat: [PL-25230]: Hiding dashboard if there is no CD license

avinashmadhwani02:


kapilkumawat86 (dismissed):

src/modules/45-projects-orgs/pages/LandingDashboardPage/LandingDashboardWelcomeView.tsx
```diff
@@ -30,15 +31,15 @@ interface WelcomeViewProps {
...
-
 const LandingDashboardWelcomeView: React.FC<WelcomeViewProps> = props => {
   const { accountId } = useParams<AccountPathProps>()
   const history = useHistory()
   const { currentUserInfo } = useAppStore()
   const { getString } = useStrings()
   const className = currentUserInfo.emailVerified === undefined || currentUserInfo.emailVerified ? '' : 'hasBanner'
   const name = currentUserInfo.name || currentUserInfo.email
+  const { CD_LICENSE_STATE } = useLicenseStore()
+  const hideBackButton = getLandingDashboardDefaultView(CD_LICENSE_STATE)
```
abhinavrastogi-harness, line 42: `hideBackButton` is a boolean.. why are we assigning the type `View` to it?

src/modules/45-projects-orgs/utils/utils.ts
```diff
@@ -122,3 +125,11 @@ export const getModuleFullLengthTitle = (module: ModuleName): keyof StringsMap =
       return 'common.purpose.cd.continuous'
   }
 }
+
+export const getLandingDashboardDefaultView = (cdLicenseState: LICENSE_STATE_VALUES) => {
```
abhinavrastogi-harness, line 129: this API is not extensible.. in a few weeks we will have other modules also which we will need to evaluate.

can we instead create a new hook, which itself includes `useLicenseStore` instead of the page using it and passing the license to this util.

abhinavrastogi-harness (changes requested):

mayankVermaHarness (approved):

abhinavrastogi-harness (approved):

PR #9003, opened 2022-05-23
fix: [CDS-36473]: fixed time filter reset issue for (home, services, overview,serviceDetail) dashboards

Sagar-Rajput7:
##### Summary:

> When the user navigates between different views (Services, Overview, Home,serviceDetail) the system doesn’t remember the time filter that was selected (within the same session), and every time the user goes back to the same view, the filter resets.

> stored key-value pair for timeRange on sessionStorage (using existing useLocalStorageHook )
> sessionStorage will clear on re-login/new-login, tab close, origin (https, port etc. ) change etc.

##### Screenshots:

https://user-images.githubusercontent.com/98325173/169822533-1ee766c7-8dc8-4d2d-b7d8-e455a844fa43.mov

https://user-images.githubusercontent.com/98325173/169822813-bde0f8aa-f344-462e-9dbf-ccef57a3cadd.mov

- Jest: `retrigger jest`
- ESLint: `retrigger eslint`
- Build: `retrigger build`
- Rebase: `trigger rebase`
- Cypress: `trigger cypress`


Sagar-Rajput7:
retrigger jest

src/modules/75-cd/components/ServiceDetails/ServiceDetailsContent/ServiceDetailsSummary.tsx
```diff
@@ -22,23 +22,35 @@ import { DeploymentsWidget } from '@cd/components/Services/DeploymentsWidget/Dep
...
+  )
+
+  const timeServiceDetail = validTimeFormat(timeRangeService)
+  timeRangeService.range[0] = timeServiceDetail.range[0]
+  timeRangeService.range[1] = timeServiceDetail.range[1]
   return (
     <Page.Body>
       <Container flex className={css.timeRangeContainer}>
-        <TimeRangeSelector timeRange={timeRange?.range} setTimeRange={setTimeRange} />
+        <TimeRangeSelector timeRange={timeRangeService?.range} setTimeRange={setTimeRangeService} />
```
AshwinBhatkal-Harness, line 48: You can choose to ignore this. But, I think serviceTimeRange conveys a better meaning or timeRangeService. When we look for a property of range, we would look at **TimeRange** instead of **Service**. However, timeRangeForService would convey it properly.

AshwinBhatkal-Harness (approved):

src/modules/75-cd/components/ServiceDetails/ServiceDetailsContent/ServiceDetailsSummary.tsx
Sagar-Rajput7: ok i will change in other PR which I will raise today itself - as that pr will also be depending on this.
Thanks for the suggestion.

satyam-harness (approved):

src/framework/utils/SecureStorage.ts
```diff
@@ -33,6 +33,7 @@ export default class SecureStorage {
     }
 
     localStorage.clear()
+    sessionStorage.clear()
```
kapilkumawat86, line 36: why this change, please add coment

kapilkumawat86 (approved):

src/framework/utils/SecureStorage.ts
Sagar-Rajput7: This is for clearing the sessionStorage on logout ( this is required because we want our time filters to reset to default when a new login session is triggered and for the rest, it should follow the default sessionStorage behavior )
(as existing sessionStorage behavior does not care about the login/logout of our site - but we want to do this from our end)

I will add the comment in the code -> in the next pr that I will be raising after this as that will be dependent on this.
will that be ok?

PR #9029, opened 2022-05-24
fix: [FFM-3447]: more spacing to accomodate trial banner

crystal-fong:
##### Summary:
When the trial banner appears on an old version of Target rules section, the footers were cut off unless the banner is dismissed.

The proper fix will be available when Flag Details Page Simplification and Rule enhancements (1513) epic is released

##### Jira Links:
https://harness.atlassian.net/browse/FFM-3447

##### Screenshots:
Old target page with banner
image

New target page with a banner does not cut off any components
image

- Jest: `retrigger jest`
- ESLint: `retrigger eslint`
- Build: `retrigger build`
- Rebase: `trigger rebase`
- Cypress: `trigger cypress`


crystal-fong:
retrigger titlecheck

chrisblakely01 (commented):
Can ya put some screenshots without the trial banner?

Also, for the new stuff, you have to make a change to the rules - this will trigger the bottom save/cancel buttons to appear, probs want to check this with/without the banner too 👍 

Last thing to check is the smaller screen size - this came from sales dude who is likely on a 13 inch mac

src/modules/75-cf/components/FlagActivation/FlagActivation.module.scss
```diff
@@ -11,7 +11,7 @@
 
 // this will be deprecated once FFM-1513 is complete
 .tabContainerHeight {
-  height: calc(100% - 140px);
+  height: calc(100% - 176px);
```
knagurski, line 14: Won't this cause issues when the banner is not present? Might be worth looking into if there is some way we can detect that the banner is displayed and selectively accommodate it

crystal-fong: It seems that when you dismiss the banner, theres a key value pair {license_banner_saved: {cf:true}} object saved in session storage. Need to figure how to access that

knagurski: It's not so much when it's been dismissed, more for the 99% case when it wasn't there to begin with.

crystal-fong: well if the banner isn't there the license_banner_saved object would be null

knagurski: What is it like when the banner is there but not dismissed?

crystal-fong: uglyyyyyy
image

crystal-fong: would it just be better to have the footer be fixed to the bottom rather than it being dependent on the the size of the container it's in?

knagurski: Unfortunately that would mean that would lead things being hidden behind the footer when they overflow.

src/modules/75-cf/pages/feature-flags-detail/targeting-rules-tab/TargetingRulesTab.tsx
```diff
@@ -46,6 +47,8 @@ const TargetingRulesTab = ({
   const { orgIdentifier, accountId: accountIdentifier, projectIdentifier } = useParams<Record<string, string>>()
   const { activeEnvironment: environmentIdentifier } = useActiveEnvironment()
 
+  const isLicenseBannerDismissed = sessionStorage.getItem('license_banner_dismissed')
```
knagurski, line 50: Unfortunately I don't think this will work. `sessionStorage` (and `localStorage`) store all values as strings, so this will either return `null` if the storage item doesn't exist, or a string. In this case when the banner is dismissed it sets a JSON-encoded string (`{"cf":true}`) stating which module it was dismissed in. Regardless, the presence or absence of this storage item doesn't tell you whether the banner is displayed, it only tells you that it was dismissed in a specific module. In an account that isn't on an enterprise trial, the banner isn't display and this storage item isn't set. In an account that is on an enterprise trial, this banner is displayed and this item isn't set.

crystal-fong: yup, just confirmed that too. Pretty stumped now, was thinking `useLicenseStore` might have something usable there but I don't know enough about it

knagurski: One option is to add an attribute in [the banner](https://github.com/harness/harness-core-ui/blob/develop/src/modules/10-common/layouts/TrialLicenseBanner.tsx#L295) and then detect it in FF. You can detect it programatically using a `useEffect`, or even better, just use CSS with something like this...

```scss
:global([data-whatever-we-used] ~ * [data-someglobalthingforourlayout]) {
  height: calc(100vh - 196px);
}
```

src/modules/10-common/layouts/TrialLicenseBanner.tsx
```diff
@@ -292,7 +292,7 @@ export const TrialLicenseBanner = (): React.ReactElement => {
   }
 
   return (
-    <div className={cx(css.trialLicenseBanner, expiredClassName)}>
+    <div className={cx(css.trialLicenseBanner, expiredClassName)} data-banner-is-visible="true">
```
knagurski, line 295: Maybe be a wee bit more specific? And you probably don't need the value
```suggestion
    <div className={cx(css.trialLicenseBanner, expiredClassName)} data-trial-license-banner-visible>
```

knagurski (dismissed):
Love when we can use CSS to fix a problem rather than JS

src/modules/10-common/layouts/TrialLicenseBanner.tsx
crystal-fong: yeah fair, mind blown when I saw it working

knagurski: The power of attribute and sibling selectors in CSS!

crystal-fong:
retrigger coverage

src/modules/75-cf/pages/feature-flags-detail/targeting-rules-tab/TargetingRulesTab.tsx
nanaxuHarness: this is magic, respect!

laura-mannus (approved):

knagurski (approved):
Nice and simple. Good work 👍🏻

PR #9034, opened 2022-05-24
feat: [FFM-2706]: Flag listing

knagurski:
This PR refactors the Target Group detail page Flag listing to extract it to a reusable component for use in both the Target Group detail page and the Target detail page. This PR also adds support for disabling the form due to RBAC and Plan Enforcement issues and fixes an overflow issue that effects the left bar.

|Before|After|
|------|------|
|Before|After|

Refs [FFM-2706](https://harness.atlassian.net/browse/FFM-2706)

### Additions
- 

### Changes
-


knagurski:
trigger cypress

nanaxuHarness (approved):

azharness (approved):

src/modules/75-cf/pages/target-detail/components/FlagSettings/FlagSettings.tsx
```diff
@@ -0,0 +1,108 @@
...
+    },
+    [initialValues, patchTarget, refetchFlags, showError]
+  )
+
+  if (loadingFlags) {
+    return <ContainerSpinner flex={{ align: 'center-center' }} />
+  }
+
+  if (flagsError) {
+    return <Page.Error message={getErrorMessage(flagsError)} onClick={() => refetchFlags()} />
```
Harness-JKhattak, line 94: In the onClick, could you call "refetchFlags" directly without the anonymous function?

```suggestion
    return <Page.Error message={getErrorMessage(flagsError)} onClick={refetchFlags} />
```

Harness-JKhattak (approved):
🦣  PR. Looks good though, I like how your module is structured and want to get ours towards that as well.

Only could spot a couple of very minor things 👍

src/modules/75-cf/pages/target-detail/components/FlagSettings/FlagSettings.tsx
knagurski: Looking at the `PageError` component, the `onClick` is just passed on directly from a `Button` component, so it'll pass the click event as the 1st argument. The `refetchFlags` function accepts an argument to override the original request params. So doing it direct would pass the click event as the override request params and who knows what it would do then 🤪

Harness-JKhattak: Only one way to find out 💥

PR #9296, opened 2022-06-07
feat: [PL-25725]: Adding Uts for GitSyncForm

sunnykesh-harness:
##### Screenshots:

- Jest: `retrigger jest`
- ESLint: `retrigger eslint`
- Build: `retrigger build`
- Rebase: `trigger rebase`
- Cypress: `trigger cypress`


src/modules/10-common/interfaces/GitSyncInterface.ts
```diff
@@ -35,3 +35,25 @@ export const Entities: Entity = {
   SERVICE: 'Service',
   TEMPLATE: 'Template'
 }
+export interface PipelineInfoConfigWithGitDetails extends PipelineInfoConfig {
```
chetanvekariya-harness, line 38: No point in keeping these in common file, if they are not used anywhere else except pipeline create.
Also, if they are used for many entities and we want to keep them in common file then we should rename these to some generic name

chetanvekariya-harness (dismissed):

src/modules/10-common/interfaces/GitSyncInterface.ts
sunnykesh-harness: ok, have not moved them now. But, these interfaces might be required in test files too.

PR #9574, opened 2022-06-21
fix: [FFM-3714]: Display delete variation buttons for multivariate flags

knagurski:
This PR ensures that the delete variation buttons are shown for multivariate flags. Previously the button was visually hidden, but still interactive only when there were 2 variations. This PR also switches to use the `FieldArray` component and helpers, moves the title to the `title` prop and adds some missing tests.

|Before|After|
|------|------|
|Before|After|

Refs [FFM-3714](https://harness.atlassian.net/browse/FFM-3714)

### Additions
- added tests

### Changes
- moved title from `Heading` tag to the `title` prop of the modal
- removed custom code to remove a variation and replaced with `remove`
- removed custom code to add a variation and replaced with `push`
- removed visual-only hiding of delete buttons
- made delete variation button accessible
- optimised and un-skipped previously flaky test


crystal-fong (approved):
Nice work! 🔥

d1wilko (approved):
looks good :)

PR #9713, opened 2022-06-28
fix: [CDS-4091]: Pipeine/deployment search filter fixed

goyalsonali:
##### Jira Links:

https://harness.atlassian.net/browse/PIE-4091

##### Screenshots:

https://user-images.githubusercontent.com/98508399/176283750-0f4a44e2-1c89-4172-9e27-c3467d3afbc9.mov

- Jest: `retrigger jest`
- ESLint: `retrigger eslint`
- Build: `retrigger build`
- Rebase: `trigger rebase`
- Cypress: `trigger cypress`


src/modules/70-pipeline/pages/pipelines/PipelinesPage.tsx
```diff
@@ -195,7 +195,7 @@ function PipelinesPage({ mockData }: CDPipelinesPageProps): React.ReactElement {
 
   function handleQueryChange(query: string): void {
     if (query) {
-      updateQueryParams({ searchTerm: query })
+      updateQueryParams({ searchTerm: query, page: defaultPageNumber.toString() })
```
udayvunnam, line 198: Why does this needs to be a string only for this specific case ?

goyalsonali: We have updateQueryParams using this :  useUpdateQueryParams<Partial<StringQueryParams>>()
Due to which we need to paas only string in this case.

udayvunnam (approved):

src/modules/70-pipeline/pages/pipelines/PipelinesPage.tsx
bhavya-sinha: I think `StringQueryParams` type is itself wrong. We should not restrict the page value to string. We are not doing this at other place. Please update  `StringQueryParams `. I think it was added by @satyam-harness

vkbansal (approved):

PR #9879, opened 2022-07-07
fix: [FFM-3984]: Manage activeEnvironment when selecting env

knagurski:
This PR moves the logic for triggering a change to the active environment into the `useEnvironmentSelectV2` hook and updates the `rewriteCurrentLocationWithActiveEnvironment` function to be able to remove the active environment.


src/modules/75-cf/utils/CFUtils.ts
```diff
@@ -272,78 +271,22 @@ export function useValidateVariationValues(): UseValidateVariationValuesResult {
...
-  '#e4e4eb',
-  '#ebebf0',
-  '#f2f2f5',
-  '#fafafc',
-  '#ffffff', // grey
-  '#1c1c28',
-  '#28293d',
-  '#555770',
-  '#8f90a6',
-  '#c7c9d9' // black
```
Kira-Jennings, line 321: Is this unused/not needed/moved to somewhere else?

knagurski: Was never used (there's a lot of stuff like that), so yanked it when I was in the neighbourhood.

crystal-fong (approved):

PR #9959, opened 2022-07-11
fix: [FFM-3982]: Hide Filter Cards when no Environment

Kira-Jennings:
Hide 'Active Flags' filter card behind feature flag until backend fixes are complete
- unit test added

177830724-77369a81-4e64-4727-b8b9-34969a4b6ea2


knagurski (commented):
Looks nice and simple 👍🏻 Can we get a test to prove it hides when no envs and shows when there are envs?

src/modules/75-cf/pages/feature-flags/__tests__/FeatureFlagsPage3.test.tsx
```diff
@@ -8,16 +8,24 @@
...
 import type { RenderColumnFlagProps } from '../FeatureFlagsPage'
 import mockFeatureFlags from './mockFeatureFlags'
 import cellMock from './data/cellMock'
 
+// eslint-disable-next-line @typescript-eslint/explicit-function-return-type
+const renderActiveEnvironmentHook = (activeEnvironment = 'foobar') => {
+  window.location.hash = `path?activeEnvironment=${activeEnvironment}`
+  return renderHook(() => useActiveEnvironment())
+}
+
```
knagurski, line 28: I'm pretty sure you can remove this and just keep the `window.location.hash` part in the test you want it in, as you're not actually using the rendering of the hook.

```typescript
test('should render when Feature Flags exist but no active Environment', async () => {
  window.location.hash = 'path?activeEnvironment=something'
  renderComponent()

  expect(screen.queryAllByTestId('filter-card').length).toBe(6)
})
```

An alternative would be to mock the `useActiveEnvironment` hook...

```typescript
import * as UseActiveEnvironmentHook from '@cf/hooks/useActiveEnvironment

// ...

describe('FeatureFlagsPage', () => {
  const useActiveEnvironmentMock = jest.spyOn(UseActiveEnvironmentHook, 'default')

  beforeEach(() => {
    useActiveEnvironmentMock.mockReturnValue({ activeEnvironment: 'something' })
  })

  // ...

  test('should not render if there is no active Environment', async () => {
    useActiveEnvironmentMock.mockReturnValue({ activeEnvironment: undefined })

    renderComponent()

    // ...
  }
  
  // ...
}
```

src/modules/75-cf/pages/feature-flags/__tests__/FeatureFlagsPage3.test.tsx
```diff
@@ -127,31 +108,38 @@ describe('FeatureFlagsPage', () => {
...
+      expect(screen.queryAllByTestId('filter-card').length).toBe(0)
+    })
+    test('should render when Feature Flags exist but no active Environment', async () => {
+      renderActiveEnvironmentHook()
+      renderComponent()
+
+      expect(screen.queryAllByTestId('filter-card').length).toBe(6)
+    })
+    test('should not render if there is an active Environment but no flags', async () => {
```
knagurski, line 132: Doesn't look like you're setting an active environment in this test

src/modules/75-cf/pages/feature-flags/__tests__/FeatureFlagsPage3.test.tsx
```diff
@@ -127,31 +108,38 @@ describe('FeatureFlagsPage', () => {
...
+      mockImport('services/cf', {
+        useGetAllFeatures: () => ({ data: undefined, refetch: jest.fn() })
+      })
+      renderComponent()
+
+      expect(screen.queryAllByTestId('filter-card').length).toBe(0)
```
knagurski, line 138: There's a nicer assertion available for lengths
```suggestion
      expect(screen.queryAllByTestId('filter-card')).toHaveLength(0)
```

src/modules/75-cf/pages/feature-flags/__tests__/FeatureFlagsPage3.test.tsx
Kira-Jennings: It's already been set in the previous

Kira-Jennings: It doesn't reset after that

knagurski: You should never rely on the state from a previous test. All tests should be able to be run independently.

Kira-Jennings: That's why these are in their own describe block, only these three tests care about the activeEnvironment

src/modules/75-cf/pages/feature-flags/__tests__/FeatureFlagsPage3.test.tsx
Kira-Jennings: This has proven to be far too problematic, breaking existing tests and typechecking etc. I've got them all working to a point but the tests for 1. clicking the row, and 2. Clicking edit I can't get working anymore. Have spent far too much time on this already :(

knagurski: This page is a pain in the butt to test. I had a wee go and opened a PR against this brach. Seems to work well, but there were a few head-scratcher moments! Feel free to take inspiration or go ahead and merge the changes. https://github.com/harness/harness-core-ui/pull/10159

knagurski: The issue you were probably seeing was due to the code that redirects when the active env was not a known env or it was empty (but envs do exist). There are some places in these tests where `fireEvents` are stacked and assertions are not waited for, so a race-condition would allow the redirect to sneak in there and swap the `activeEnvironment` out from under you! This would be triggered for places where an `activeEnvironment` wasn't specific (in those cases, it would redirect to the first one). And then there's a different mocked location in the `mockEnvironments` mocks and the `mockFeatureFlags` mocks. Really threw me when I wasn't expecting `sfgsd` and `foobar`.

Would be lovely to completely rethink the tests on this page... some day, but not today ¯\_(ツ)_/¯

knagurski (approved):
Nice work 👍🏻

chrisblakely01 (approved):

PR #10103, opened 2022-07-19
chore: [CDB-275]: Custom Dashboards - Fixed ESLint warnings

Harness-JKhattak:
### Summary

- Chore to fix all ESLint warnings within the Dashboards module without changing/effecting functionality 

#### PR Checklist

<details>

</details>

<details>

- Jest: `retrigger jest`
- ESLint: `retrigger eslint`
- Build: `retrigger build`
- Rebase: `trigger rebase`
- Cypress: `trigger cypress`
</details>


src/modules/45-dashboards/components/DashboardList/DashboardList.tsx
```diff
@@ -37,6 +37,7 @@ const DashboardList: React.FC<DashboardListProps> = ({
   const { getString } = useStrings()
   const { accountId, folderId } = useParams<{ accountId: string; folderId: string }>()
 
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
   type CustomColumn<T extends Record<string, any>> = Column<T>
```
knagurski, line 41: Another way, instead of using `any`, would be to use `unknown`. You'll then have to assert what's in it when you go to use it, but it's more type-safe. Another option might be to allow another type to be passed in which then asserts what the `any` is.

Not a blocker, just an observation.

src/modules/45-dashboards/components/ModuleTagsFilter/ModuleTagsFilter.tsx
```diff
@@ -14,14 +14,19 @@ import moduleTagCss from '@dashboards/common/ModuleTags.module.scss'
...
-  setPredefinedFilter: (moduleName: string, checked: boolean) => void
+  setPredefinedFilter: (moduleName: DashboardTags, checked: boolean) => void
 }
 
 const ModuleTagsFilter: React.FC<ModuleTagsFilterProps> = ({ selectedFilter, setPredefinedFilter }) => {
   const { getString } = useStrings()
   const { CENG_ENABLED, CING_ENABLED, CDNG_ENABLED, CFNG_ENABLED, CUSTOM_DASHBOARD_V2, SECURITY } = useFeatureFlags()
 
-  const renderTagsFilter = (moduleName: DashboardTags, cssClass: string, text: StringKeys, isEnabled = false) => {
+  const renderTagsFilter = (
```
knagurski, line 24: Beyond the scope of this ticket, but this essentially is a component, so it really should be in its own file with its own tests.

src/modules/45-dashboards/components/DashboardList/DashboardList.tsx
Harness-JKhattak: I will give that a go. 

I tried using the "DashboardModel" object but I'd need to manually update the object within the shared Service file to be something like below which isn't really possible with the auto-generated code 😢

```
interface DashboardModel {
  [key: string]: string | number;
  myVal: string;
}
```

src/modules/45-dashboards/components/WidgetLibrary/WidgetLibrary.tsx
```diff
@@ -62,7 +64,7 @@ export const WizardLibrary: React.FC<{ onClose: () => void }> = ({ onClose }): J
     setLibrary(filteredResults)
   }
 
-  const onSearch = (searchData: string) => {
+  const onSearch = (searchData: string): void => {
```
knagurski, line 67: Totally beyond the scope of this ticket, but for future reference, functions like this can cause unnecessary rerenders of child components as they are redefined on each render of the component. Try using `useCallback` in those cases... it's essentially `useMemo`, but for functions.

src/modules/45-dashboards/components/WidgetLibrary/WidgetLibrary.tsx
```diff
@@ -114,7 +116,7 @@ export const WizardLibrary: React.FC<{ onClose: () => void }> = ({ onClose }): J
               )}
               {library
                 ?.filter(x => x.module === MODULE_TYPES.CE)
-                .map((widget: any) => {
+                .map((widget: Library) => {
```
knagurski, line 119: You might not need the type here at all. `library`'s typing as `Library[]` should carry though automatically to the `map`. Sometimes TS does get annoyed without it, so worth a try with and without.

src/modules/45-dashboards/components/WidgetLibrary/WidgetLibrary.tsx
Harness-JKhattak: Good catch

src/modules/45-dashboards/components/WidgetLibrary/WidgetLibrary.tsx
```diff
@@ -155,12 +157,7 @@ export const WizardLibrary: React.FC<{ onClose: () => void }> = ({ onClose }): J
           )}
 
           <Layout.Horizontal className={css.footerActions}>
-            <Button
-              text="Add to Dashboard"
-              intent="primary"
-              disabled={!selectedLook}
-              onClick={() => addWidgetToDashboard()}
-            />
+            <Button text="Add to Dashboard" intent="primary" disabled={!selectedLook} onClick={addWidgetToDashboard} />
```
knagurski, line 160: You'll need a `ButtonVariation` here as well.
```suggestion
            <Button
              text="Add to Dashboard"
              intent="primary"
              variation={ButtonVariation.PRIMARY}
              disabled={!selectedLook}
              onClick={addWidgetToDashboard}
            />
```

src/modules/45-dashboards/pages/DashboardsCommonHeaderPage.tsx
```diff
@@ -26,15 +26,18 @@ const DashboardsHeader: React.FC = () => {
   const { accountId, folderId } = useParams<{ accountId: string; folderId: string }>()
   const [isOpen, setDrawerOpen] = useState(false)
 
-  const title = breadcrumbs.length ? breadcrumbs[breadcrumbs.length - 1].label : getString('common.dashboards')
+  const title = React.useMemo(
+    () => (breadcrumbs.length ? breadcrumbs[breadcrumbs.length - 1].label : getString('common.dashboards')),
```
knagurski, line 30: Always hate how there's no elegant way in JS to get the first/last element in an array. Other languages have `.first` and `.last` accessors. ES2022 does have the `.at()` function, so you can do `.breadcrumbs.at(-1)` to get the last one, but we don't use ES2022 yet. This is totally fine though, just thinking outloud.

src/modules/45-dashboards/pages/DashboardsCommonHeaderPage.tsx
```diff
@@ -26,15 +26,18 @@ const DashboardsHeader: React.FC = () => {
...
+    [breadcrumbs, getString]
+  )
 
   React.useEffect(() => {
     const titleArray: string[] = [getString('common.dashboards')]
 
     if (breadcrumbs) breadcrumbs.map(breadcrumb => titleArray.push(breadcrumb.label))
```
knagurski, line 37: Since you don't care about the return of the `map` function, you should probably use `forEach`
```suggestion
    if (breadcrumbs) {
      breadcrumbs.forEach(({ label }) => titleArray.push(label))
    }
```

src/modules/45-dashboards/pages/dashboardView/DashboardView.tsx
```diff
@@ -70,6 +70,7 @@ const DashboardViewPage: React.FC = () => {
     return () => {
       window.removeEventListener('message', lookerEventHandler)
     }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [])
```
knagurski, line 74: Not sure if this applies to Dashboards (is this specific to the project?), but have you tested swapping projects from the sidebar when already on this page? If it is project-specific, you might have a bug lurking. If it isn't project-specific, you're fine 👍🏻

src/modules/45-dashboards/pages/home/DashboardForm.tsx
```diff
@@ -103,6 +103,7 @@ const DashboardForm: React.FC<DashboardFormProps> = ({
       name: formData?.name || '',
       description: formData?.description?.split(TAGS_SEPARATOR) || []
     }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
knagurski, line 106: As-is the return of this `useMemo` won't change in relation to the `formData` changing. Is that intentional?

src/modules/45-dashboards/pages/dashboardView/DashboardView.tsx
Harness-JKhattak: This event handler is specific to the embedded looker app itself so resetting them when a project/view changes shouldn't make a difference

knagurski: Issue right now may be that it won't reset on a project change, regardless of what the embed is doing

src/modules/45-dashboards/components/ModuleTagsFilter/ModuleTagsFilter.tsx
Harness-JKhattak: Created a tech ticket to address this in multiple components -> https://harness.atlassian.net/browse/CDB-276

src/modules/45-dashboards/pages/dashboardView/DashboardView.tsx
Harness-JKhattak: Dashboards doesn't have the ability to switch projects from the sidebar. What I'll do is add in the dependency for the dashboard "viewId" so that if the user switches dashboards, it'll reset the event listener 👍

knagurski (approved):
🔥

laura-mannus (approved):
Nice one :)

Harness-JKhattak:
retrigger cypress

Harness-JKhattak:
trigger cypress

PR #10159, opened 2022-07-21
fix: [FFM-3982]: Mock envs for testing

knagurski:
This PR adjusts the mocking of envs in the Feature Flags listing page tests.


src/modules/75-cf/pages/feature-flags/__tests__/FeatureFlagsPage3.test.tsx
```diff
@@ -53,23 +46,36 @@ const renderComponent = (): RenderResult =>
...
+    })
+  })
+
+  window.location.hash = newLocation
+}
+
 describe('FeatureFlagsPage', () => {
-  beforeAll(() => {
+  beforeEach(() => {
+    jest.clearAllMocks()
```
knagurski, line 72: This might not be needed at all... all it does is reset the call info around existing mocks (i.e. how want times the mock was called and with what arguments). If you want to clear down mocks, then you need `jest.resetAllMocks` instead, but you don't really need to do that manually as Jest isolates each test suite and mocks set up in the suite are always killed off.

src/modules/75-cf/pages/feature-flags/__tests__/FeatureFlagsPage3.test.tsx
```diff
@@ -119,23 +127,27 @@ describe('FeatureFlagsPage', () => {
...
+      mockEnvs(false)
       renderComponent()
 
-      expect(screen.queryAllByTestId('filter-card').length).toBe(0)
+      expect(screen.queryAllByTestId('filter-card')).toHaveLength(0)
     })
-    test('should render when Feature Flags exist but no active Environment', async () => {
-      renderActiveEnvironmentHook()
+
+    test('should render when Feature Flags exist and there is an active Environment', async () => {
```
knagurski, line 136: I changed this one to include an active env as the code only requires an active env to render.

src/modules/75-cf/pages/feature-flags/__tests__/FeatureFlagsPage3.test.tsx
```diff
@@ -119,23 +127,27 @@ describe('FeatureFlagsPage', () => {
...
+      mockEnvs(true)
```
knagurski, line 137: `mockEnvs(true)` isn't really required since it's being done in the `beforeEach`, but it's a good idea to be super-blunt about what your setup is in a test, just to remove any ambiguity when reading the test.

src/modules/75-cf/pages/feature-flags/__tests__/FeatureFlagsPage3.test.tsx
```diff
@@ -53,23 +46,36 @@ const renderComponent = (): RenderResult =>
...
+    useGetEnvironmentListForProject: () => ({
+      data,
+      loading: false,
+      error: undefined,
+      refetch: jest.fn()
+    })
+  })
+
+  window.location.hash = newLocation
+}
```
knagurski, line 68: What we're doing here is making sure the environments mock and location are controlled together so we don't accidentally trip the auto-redirect code and get unexpected results.

PR #10168, opened 2022-07-22
feat: [SRM-11182]: UX review comments

deepesh-ui:
### Summary.  UX changes as per https://harness.atlassian.net/browse/SRM-11182

Updated Title container for Templates card and Template studio  from SVG to  Styled component

#### Screenshots
Screenshot 2022-07-26 at 12 21 12 PM
Screenshot 2022-07-26 at 12 20 09 PM

NA

#### PR Checklist

<details>

</details>

<details>

- Jest: `retrigger jest`
- ESLint: `retrigger eslint`
- Build: `retrigger build`
- Rebase: `trigger rebase`
- Cypress: `trigger cypress`
</details>


src/framework/Templates/TemplateConfigModal/TemplateConfigModal.tsx
```diff
@@ -92,6 +93,18 @@ export interface ConfigModalProps extends ModalProps {
   onClose: () => void
 }
 
+export const getTemplateTitleFromType = (
```
deepesh-ui, line 96: Inside **Create New Template**  Modal, we don't want  to populate the default new template name rather we are keeping it Empty.

yc147 (commented):
When creating a new template, you have removed the default value for the name field i.e. New {entity} Template Name in case of monitored service template. If this is a real use case then let’s solve it in a way that in future other templates can also benefit.

Also, in case of monitored service template you are getting template type label from a separate string in the codebase instead of using template.type.

yc147 (changes requested):
When creating a new template, you have removed the default value for the name field i.e. New {entity} Template Name in case of monitored service template. If this is a real use case then let’s solve it in a way that in future other templates can also benefit.

Also, in case of monitored service template you are getting template type label from a separate string in the codebase instead of using template.type.

yc147 (approved):

praneshg239 (approved):

deepesh-ui:
Coverage failure is due to newly added templates , which don't have any functionality only classes are added for future reference

Screenshot 2022-07-26 at 4 09 28 PM

PR #10221, opened 2022-07-26
fix: [PL-26302]: add mutex for login redirect

abhinavrastogi-harness:
### Summary

Root cause:
1. We make two parallel API calls when the app starts up: getFeatureFlags and  getCurrentUserInfo
2. When a token is present, but invalid (expired), in prod, both the APIs return a 401
on receiving 401, the global response handler triggers for both errors, and tried to redirect to /auth twice. this results in the first redirect getting cancelled.
3. the redirect to /auth is via a intermediate page /redirect?returnUrl={currentUrl}
because of one redirect getting cancelled, somehow the second redirect picks the returnUrl of the current page, which is already a redirect to sign-in.
4. this results in the double returnUrl in the URL, which sends the user back to login after a successful login

Other observations:
1. this doesn’t always happen because it depends on the timings of the network calls and the dns cache status of the browser. 
2. If there is no dns cache, the second api call takes longer, and gets cancelled before it starts
3. if there is a cache, the second api call starts at the same time, and both get a separate 401

This was not happening in local because of a different timing and presence of a network proxy which adds time to both calls.

The fix is to trigger the 401 redirect only once. 

#### PR Checklist

<details>

</details>

<details>

- Jest: `retrigger jest`
- ESLint: `retrigger eslint`
- Build: `retrigger build`
- Rebase: `trigger rebase`
- Cypress: `trigger cypress`
</details>


abhinavrastogi-harness:
retrigger titlecheck

mayankVermaHarness (dismissed):

src/framework/app/App.tsx
```diff
@@ -132,28 +133,13 @@ export function AppWithAuthentication(props: AppProps): React.ReactElement {
   Harness.openTooltipEditor = () => setShowTooltipEditor(true)
 
   const globalResponseHandler = (response: Response): void => {
-    const token = SessionToken.getToken()
     if (!response.ok) {
       switch (response.status) {
         case 401: {
-          if (token) {
-            const lastTokenSetTime = SessionToken.getLastTokenSetTime() as number
-            window.bugsnagClient?.notify?.(
```
mayankVermaHarness, line 141: Do we not want to log this anymore?

upadhyay-an (approved):

src/framework/app/App.tsx
abhinavrastogi-harness: no. for a few reasons:
1. login issues are mostly resolved. we don't need this noisy notification anymore
2. it's hard to differentiate b/w genuine 401s and wrong 401s, so this is not very helpful
3. it doesn't always fire reliably. if the navigation happens before this completes, it gets cancelled.

avinashmadhwani02 (dismissed):

abhinavrastogi-harness:
retrigger jest

abhinavrastogi-harness:
retrigger coverage

mayankVermaHarness (dismissed):

src/microfrontends/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@harness/microfrontends",
-  "version": "1.1.1",
+  "version": "1.2.1",
```
jenil-harness, line 3:
```suggestion
  "version": "1.2.0",
```

avinashmadhwani02 (approved):

mayankVermaHarness (approved):

src/framework/utils/SessionUtils.ts
```diff
@@ -26,3 +31,25 @@ export const getForgotPasswordURL = (): string => {
...
     : `${window.location.pathname.replace(/\/ng\//, '/')}#/forgot-password`
 }
+
+export interface UseLogoutReturn {
+  forceLogout: () => void
+}
+
+export const useLogout = (): UseLogoutReturn => {
+  const history = useHistory()
+  let isTriggered = false
```
knagurski, line 41: I think you may need to move this outside of the hook, otherwise if the hook is called more than once (which will be the case if the hook itself is passed to child apps, rather than the `forceLogout` function), then each invocation will start with `isTriggered` as `false`. By moving it out of the hook, all places `useLogout` is used will share the same variable.

knagurski: Alternatively, instead of passing `useLogout` as a hook, pass the resultant `forceLogout` as a standard prop.

src/framework/utils/__tests__/SessionUtils.test.ts
```diff
@@ -5,7 +5,18 @@
  * https://polyformproject.org/wp-content/uploads/2020/06/PolyForm-Shield-1.0.0.txt.
  */
 
-import { getLoginPageURL } from '../SessionUtils'
+import { renderHook } from '@testing-library/react-hooks'
+import { act } from 'react-dom/test-utils'
```
knagurski, line 9: You should grab `act` from `@testing-library/react-hooks` instead
```suggestion
import { act, renderHook } from '@testing-library/react-hooks'
```

knagurski (commented):
Some suggestions and a query about the use of a scoped variable to track whether the logout was called.

src/framework/utils/SessionUtils.ts
abhinavrastogi-harness: @knagurski good catch! but is this a blocker? currently no MFEs will be using the new hook. the rest of the code changes fix a production bug. is it okay if I refactor according to your comments in a separate PR next week?

knagurski: Looks like this PR passes it specifically to the Governance App. As-is, practically speaking, you'd need to hit a 401 in both the parent and child app at the same time for it to be an issue, which probably isn't likely (the parent will probably have triggered a 401 well before the child, or it'll be just from the child). Worth fixing, but deffo not a blocker 👍🏻

knagurski (approved):
Nice 👍🏻

PR #10355, opened 2022-08-02
feat: [PIE-4673]: handle execution input type in components

vkbansal:
### Summary

We have added a new type of runtime input which is of type `EXECUTION_TIME`. Before this change we were just checking the type to directly like `type === MultiTypeInputType.RUNTIME`. This will not work correctly once the new type is enabled.

This PR handles this case and also add an ESLint rule to avoid such use cases and add a helper method `isMultiTypeRuntime` which should be used instead.

`getMultiTypeFromValue(value) === MultiTypeInputType.RUNTIME` is still valid as it handles the type properly and only returns `MultiTypeInputType.RUNTIME` and never `MultiTypeInputType.EXECUTION_TIME` . The issue only comes when second parameter `allowedTypes`  is passed. It overrides the return types and returns one of the values from the given array and this array may or may not contain `MultiTypeInputType.EXECUTION_TIME`. But we have to handle even this case. Hence only cases where the second param is passed is flagged by the rule.

The ESLint rule works as follows:

```
// CORRECT USAGE:
getMultiTypeFromValue(value) === MultiTypeInputType.RUNTIME
isMultiTypeRuntime(type)

// WRONG USAGE:
 someType === MultiTypeInputType.RUNTIME
 getMultiTypeFromValue(value, allowedTypes) === MultiTypeInputType.RUNTIME // must not be used with allowedTypes
```

#### Screenshots

NA

#### PR Checklist

<details>

</details>

<details>

- Jest: `retrigger jest`
- ESLint: `retrigger eslint`
- Build: `retrigger build`
- Rebase: `trigger rebase`
- Cypress: `trigger cypress`
</details>


vkbansal:
All the sonarqube issues are existing ones and should not be fixed in this PR.

bhavya-sinha (approved):

kapilkumawat86 (approved):

PR #10799, opened 2022-08-25
fix: [FFM-4345]: update form fields to use TagInput only when 'in' operator selected

Kira-Jennings:
- Form fields in section 'Target Based on Condition' have been updated to allow multiple values only when the 'in' operator has been selected
- updated unit tests

image
image


diptiman-adak:
 


            
 #### 3 Issues 
* Code issue 3 Code issues 

 [Details](https://app.embold.io/organization/gh/harness/repositories/e864badd0b95ab81c060b15b85f15701/pull-request/10799)

crystal-fong (dismissed):

src/modules/75-cf/pages/target-group-detail/components/EditTargetGroupCriteria/__tests__/RuleRow.test.tsx
```diff
@@ -81,9 +90,74 @@ describe('RuleRow', () => {
     })
   })
 
-  test('it should display a tag input for the value', async () => {
+  test('it should display a text input for the value when no op selected', async () => {
     renderComponent()
-    expect(screen.getByLabelText('cf.segmentDetail.values')).toBeInTheDocument()
+
+    const ruleRow = screen.getByTestId('ruleRowFields')
```
laura-mannus, line 96: you can move this outside the tests and declare it only once so you don't have to remake it in every test

Kira-Jennings: That's usually the case but in this scenario that would mean declaring it outside of the scope of the test, and therefore before anything is rendered which results in ruleRow being undefined
If we didn't need to send a different operator every time we're calling `renderComponent` I could have called the render within a beforeEach, then declared `ruleRow` outside of the tests with no problems

Kira-Jennings: image
image

Kira-Jennings: I've removed all the `waitFor` blocks in this file that weren't needed, only the one in last test is required

laura-mannus (approved):
nice

crystal-fong (approved):

crystal-fong (approved):

PR #10932, opened 2022-09-01
fix: [cds-42447]: set empty inputs on infra switch

azharness:
### Summary

When infrastructure is a runtime input, when user tries to run pipeline, and switch between different infras from dropdown, infra runtime inputs should be updated.
If select one infra with runtime inputs, this fields are shown but when select another one which don't contain runtime inputs, previous fields have not been removed.

#### Screenshots

NA

#### PR Checklist

<details>

</details>

<details>

- Jest: `retrigger jest`
- ESLint: `retrigger eslint`
- Build: `retrigger build`
- Rebase: `trigger rebase`
- Cypress: `trigger cypress`
</details>


Tamaraius (approved):

src/modules/75-cd/components/PipelineSteps/DeployInfrastructureStep/DeployInfrastructures/DeployInfrastructures.tsx
```diff
@@ -130,6 +130,19 @@ function DeployInfrastructures({
         } else {
           formik?.setFieldValue('infrastructureInputs', parsedInfrastructureDefinitionYaml)
         }
+      } else {
+        const selInfraParsed = parse(defaultTo(selectedInfrastructure, ''))
```
AshwinBhatkal-Harness, line 134: This should defaultTo '{}'. Parsing '' would throw an error

AshwinBhatkal-Harness: Also, can we have a complete name for this please

src/modules/75-cd/components/PipelineSteps/DeployInfrastructureStep/DeployInfrastructures/DeployInfrastructures.tsx
```diff
@@ -130,6 +130,19 @@ function DeployInfrastructures({
...
+        if (selInfraParsed) {
+          const selectedInfraWithEmptyInputs = {
+            identifier: selInfraParsed.infrastructureDefinition.identifier,
+            inputs: {
```
AshwinBhatkal-Harness, line 138: We can remove the inputs key completely. Just the identifier would do

azharness: Actually we need to set type, otherwise will say that step was not found, but will remove inside identifier and spec, it works without them

src/modules/75-cd/components/PipelineSteps/DeployInfrastructureStep/DeployInfrastructures/DeployInfrastructures.tsx
azharness: Have checked, parse('') will return null, without throwing error, but have updated condition a bit better

src/modules/75-cd/components/PipelineSteps/DeployInfrastructureStep/DeployInfrastructures/DeployInfrastructures.tsx
AshwinBhatkal-Harness: It would say step not found if we are not updating formik values. Can we update the formik value as well as template? It does not make sense to update the template with a value that is not required

AshwinBhatkal-Harness:
Does this work on pipeline studio?

src/modules/75-cd/components/PipelineSteps/DeployInfrastructureStep/DeployInfrastructures/DeployInfrastructures.tsx
AshwinBhatkal-Harness: Also, can we do the above only if path exists? We do not want to change anything on the pipeline studio end

azharness: Yes, you are right, formik updated also :)

azharness:
retrigger jest

azharness:
trigger cypress

AshwinBhatkal-Harness (approved):

src/modules/75-cd/components/PipelineSteps/DeployInfrastructureStep/DeployInfrastructures/DeployInfrastructures.tsx
AshwinBhatkal-Harness: About the formik update, I meant to remove the "inputs" key and then update the corresponding formik value. This way, you will not get the error of step not found.
Inputs as a key is technically not correct if there exists no runtime inputs

Tamaraius (approved):

PR #11089, opened 2022-09-07
feat: [FFM-4313]: Redirect to SRM page button

crystal-fong:
### Summary
If users do not have any existing Services in their project, the empty state of Services dialog should render a button that will take them to the SRM module where they can add a new Service.

If users search for a non-existing Service, the same empty state should appear but without the `Add New Service` button

#### Screenshots

NA

https://user-images.githubusercontent.com/95680496/188860301-54aa2a86-8a24-4efc-bcd9-215b417fad15.mov

New empty states with an image

image

image

#### PR Checklist

<details>

</details>

<details>

- Jest: `retrigger jest`
- ESLint: `retrigger eslint`
- Build: `retrigger build`
- Rebase: `trigger rebase`
- Cypress: `trigger cypress`
</details>

## [Contributor license agreement](https://github.com/harness/harness-core-ui/blob/develop/CONTRIBUTOR_LICENSE_AGREEMENT.md)


diptiman-adak:
 


            
 #### 5 Issues 
* Code issue 5 Code issues 

 [Details](https://app.embold.io/organization/gh/harness/repositories/e864badd0b95ab81c060b15b85f15701/pull-request/11089)

laura-mannus (dismissed):
Nice!

src/modules/75-cf/components/FlagActivation/ServicesList.tsx
```diff
@@ -71,11 +81,15 @@ const EditServicesModal: FC<EditServicesProps> = ({
...
+  searchTerm,
   serviceError,
-  refetchServices
+  refetchServices,
+  queryParams
 }) => {
   const { getString } = useStrings()
-  const noServices = Boolean(!filteredServices?.length && !loading)
+
+  const noServices = Boolean(!filteredServices?.length && !loading && !searchTerm)
```
knagurski, line 91: You don't need to pass through `Boolean` as the result of the condition is already a boolean. Same below

src/modules/75-cf/components/FlagActivation/ServicesList.tsx
```diff
@@ -100,25 +114,47 @@ const EditServicesModal: FC<EditServicesProps> = ({
...
       {loading && <ContainerSpinner height="100%" margin="0" flex={{ align: 'center-center' }} />}
 
-      {noServices && (
+      {isEmptyState && (
         <Container height="100%" flex={{ align: 'center-center' }}>
-          <Text color={Color.GREY_600} font={{ variation: FontVariation.H4, weight: 'light' }}>
-            {getString('cf.featureFlagDetail.noServices')}
-          </Text>
+          <Layout.Vertical>
+            <Layout.Horizontal margin={{ bottom: 'xxxlarge' }}>
```
knagurski, line 138: You probably don't need this `Layout` as it only has a single child.

src/modules/75-cf/components/FlagActivation/ServicesList.tsx
```diff
@@ -100,25 +114,47 @@ const EditServicesModal: FC<EditServicesProps> = ({
...
-        <ServicesFooter loading={loading} onSave={onSave} onClose={closeModal} paginationProps={paginationProps} />
+        !noServices && (
+          <ServicesFooter loading={loading} onSave={onSave} onClose={closeModal} paginationProps={paginationProps} />
+        )
       }
     >
       {loading && <ContainerSpinner height="100%" margin="0" flex={{ align: 'center-center' }} />}
 
-      {noServices && (
+      {isEmptyState && (
```
knagurski, line 135: Could we use a `NoData` component instead of laying it out in here?

src/modules/75-cf/components/FlagActivation/ServicesList.tsx
knagurski, line 138: Instead of using a `margin` to push it away from it's sibling, you could try the `spacing` prop of the parent `Layout`

src/modules/75-cf/components/FlagActivation/ServicesList.tsx
crystal-fong: the empty state here doesn't have an image, which is a required prop of `NoData`

src/modules/75-cf/components/FlagActivation/ServicesList.tsx
```diff
@@ -100,25 +114,47 @@ const EditServicesModal: FC<EditServicesProps> = ({
...
+            </Layout.Horizontal>
+            {noServices && (
+              <Layout.Horizontal flex={{ align: 'center-center' }}>
+                <Link
+                  to={routes.toCVAddMonitoringServicesSetup({
+                    accountId: queryParams.accountIdentifier,
+                    ...queryParams
+                  })}
+                >
+                  <Button icon="plus" intent="primary" variation={ButtonVariation.PRIMARY}>
```
knagurski, line 151: Probably best to avoid having a `Button` inside of a `Link`... you could set the `onClick` handler of the `Button` to navigate to the route and drop the `Link` altogether.

src/modules/75-cf/components/FlagActivation/__tests__/ServicesList.test.tsx
```diff
@@ -349,4 +348,28 @@ describe('EditServicesModal', () => {
...
+    const addNewServiceButton = screen.getByRole('button', { name: 'plus newService' })
+
+    // assert correct empty state
+    await waitFor(() => {
+      expect(screen.getByText('cf.featureFlagDetail.noServices')).toBeInTheDocument()
+      expect(addNewServiceButton).toBeInTheDocument()
+
+      expect(screen.getByRole('link', { name: 'plus newService' }).getAttribute('href')).toContain(
+        '/account/dummy/cv/orgs/dummy/projects/dummy/monitoringservices/setup'
+      )
```
knagurski, line 372: You might be able to use `.toHaveAttribute` here
```suggestion
      expect(screen.getByRole('link', { name: 'plus newService' })).toHaveAttribute('href', '/account/dummy/cv/orgs/dummy/projects/dummy/monitoringservices/setup')
```

src/modules/75-cf/components/FlagActivation/ServicesList.tsx
crystal-fong: I'll ask UX if they can provide an image quickly, if so I can change it to use `NoData`

src/modules/75-cf/components/FlagActivation/__tests__/ServicesList.test.tsx
crystal-fong: is the benefit just shorter syntax?

src/modules/75-cf/components/FlagActivation/ServicesList.tsx
knagurski: If they can't, we can make the image prop optional and conditionally show it 👍🏻

src/modules/75-cf/components/FlagActivation/__tests__/ServicesList.test.tsx
knagurski: Does exactly the same thing, so don't feel the need to change. Just reads better than getting a string and asserting against it.

src/modules/75-cf/components/FlagActivation/ServicesList.tsx
crystal-fong: @knagurski 
this is the alternative I'm trying to implement atm
`  <Button
              onClick={() =>
                history.push(
                  routes.toCVAddMonitoringServicesSetup({
                    accountId: queryParams.accountIdentifier,
                    ...queryParams
                  })
                )
              }
            />`

however this breaks the unit test as it can no longer find a href associated to that button. Is it correct to use `history.push` here?

knagurski: If you are using the `TestWrapper`, when you navigate to a new URL, the wrapper will output the new URL in the body of the output; so instead of testing that the link has a specific URL, you click the link and assert that the expected URL is in the output. Kinda counter-intuitive, but the alternative is to mock `history.push` and that's specifically not allowed by eslint rules.

The criteria I typically use for whether to use a button or a link is whether the user clicks the thing expecting an action to occur, or to get to another part of the app. In this case they click the button to create something, so it's an action.

What we really want is a link that is styled like a button, but that's out of scope I think.

src/modules/75-cf/components/FlagActivation/__tests__/ServicesList.test.tsx
crystal-fong: noted for next time ✌️

src/modules/75-cf/components/FlagActivation/ServicesList.tsx
crystal-fong: Replaced with the `NoData` so it is gone

src/modules/75-cf/components/FlagActivation/ServicesList.tsx
crystal-fong: I got my pic!

src/modules/75-cf/components/FlagActivation/ServicesList.tsx
crystal-fong: Replaced with the `NoData` so it's gone

src/modules/75-cf/components/FlagActivation/ServicesList.tsx
crystal-fong: removed

src/modules/75-cf/components/FlagActivation/ServicesList.tsx
crystal-fong: > If you are using the TestWrapper, when you navigate to a new URL, the wrapper will output the new URL in the body of the output; so instead of testing that the link has a specific URL, you click the link and assert that the expected URL is in the output

That helped a lot, knowing what to expect. Updated the test

crystal-fong:
retrigger jest

src/modules/75-cf/components/FlagActivation/ServicesList.tsx
```diff
@@ -100,25 +116,40 @@ const EditServicesModal: FC<EditServicesProps> = ({
...
-        <Container height="100%" flex={{ align: 'center-center' }}>
-          <Text color={Color.GREY_600} font={{ variation: FontVariation.H4, weight: 'light' }}>
-            {getString('cf.featureFlagDetail.noServices')}
-          </Text>
+      {isEmptyState && (
+        <Container flex={{ justifyContent: 'center' }} padding="xxxlarge">
+          <NoData
+            imageURL={NoServices}
+            buttonProps={{ icon: 'plus' }}
+            buttonText={noServices ? getString('newService') : undefined}
```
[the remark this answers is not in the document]
crystal-fong: yep, the button is hidden when there are services but you are not getting any searched results back. See the first screenshot I added ^^

src/modules/75-cf/components/FlagActivation/__tests__/ServicesList.test.tsx
```diff
@@ -349,4 +348,45 @@ describe('EditServicesModal', () => {
...
+    } as any)
+
+    renderComponent()
+    userEvent.click(screen.getByRole('button', { name: 'edit-services' }))
+
+    await waitFor(() => {
+      expect(screen.getByRole('button', { name: 'plus newService' })).toBeInTheDocument()
+
+      userEvent.click(screen.getByRole('button', { name: 'plus newService' }))
+    })
```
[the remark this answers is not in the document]
crystal-fong: huh interesting how that test wasn't stuck running because that's usually what happens whenever I have an event within a waitFor so I would've caught that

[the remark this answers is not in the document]
crystal-fong: I've updated it to the fancy version

src/modules/75-cf/components/FlagActivation/__tests__/ServicesList.test.tsx
```diff
@@ -349,4 +348,45 @@ describe('EditServicesModal', () => {
...
+      refetch: jest.fn(),
+      error: false
+    } as any)
+
+    renderComponent()
+    userEvent.click(screen.getByRole('button', { name: 'edit-services' }))
+
+    await waitFor(() => {
+      expect(screen.getByRole('button', { name: 'plus newService' })).toBeInTheDocument()
+    })
```
knagurski, line 384: Now that you're using the fancy version (`findByRole`), you can actually drop this `waitFor` altogether

crystal-fong: doubly fancy

crystal-fong: the other getByRoles can stay unfancy

src/modules/75-cf/components/FlagActivation/__tests__/ServicesList.test.tsx
```diff
@@ -349,4 +348,43 @@ describe('EditServicesModal', () => {
...
+    } as any)
+
+    renderComponent()
+
+    userEvent.click(screen.getByRole('button', { name: 'edit-services' }))
+
+    await waitFor(() => {
+      expect(screen.getByRole('heading', { name: 'cf.featureFlagDetail.noServices' })).toBeInTheDocument()
+      expect(screen.getByTestId('nodata-image')).toBeInTheDocument()
+      expect(screen.getByRole('button', { name: 'plus newService' })).toBeInTheDocument()
```
Kira-Jennings, line 367: I think the checks here should be `.toBeVisible()` as an element can be in the document but not visible

Kira-Jennings (approved):

src/modules/75-cf/components/FlagActivation/__tests__/ServicesList.test.tsx
```diff
@@ -349,4 +348,43 @@ describe('EditServicesModal', () => {
...
+      loading: false,
+      data: [],
+      refetch: jest.fn(),
+      error: false
+    } as any)
+
+    renderComponent()
+    userEvent.click(screen.getByRole('button', { name: 'edit-services' }))
+
+    expect(await screen.findByRole('button', { name: 'plus newService' })).toBeInTheDocument()
```
knagurski, line 382: Sorry, communicating through GH is so tough sometimes. You can drop this line altogether, the `await screen.findByRole` on line 384 is already handling this assertion for you.

src/modules/75-cf/components/FlagActivation/__tests__/ServicesList.test.tsx
knagurski: I'm happy either way. I prefer `toBeInTheDocument` as `toBeVisible` doesn't actually do what it pretends to do... since Jest doesn't use a real browser and doesn't understand CSS, the criteria it uses to determine if something is "visible" doesn't apply to our codebase (i.e. we don't use the `hidden` attribute or inline `display: none` etc).

The the end, for our codebase, they'll act exactly the same, so it's a preference thing. I think they both read fine and both convey what we are trying to prove; my only concern is that `isVisible` kinda gives you a false sense of what it can do.

src/modules/75-cf/components/FlagActivation/ServicesList.tsx
crystal-fong: I've tried passing an empty string before but it shows a big outline

src/modules/75-cf/components/FlagActivation/__tests__/ServicesList.test.tsx
crystal-fong: I was wondering what the difference between those were! TY for explaining @knagurski

src/modules/75-cf/components/FlagActivation/__tests__/ServicesList.test.tsx
crystal-fong: sure what's one more check gonna do. r e m o v e d

crystal-fong:
retrigger jest

crystal-fong:
retrigger jest

src/modules/75-cf/components/FlagActivation/__tests__/ServicesList.test.tsx
knagurski: Seriously though, both are fine, go with whichever you think reads better 👍🏻

knagurski (approved):
/giphy awesome sauce

Kira-Jennings (approved):

PR #11162, opened 2022-09-09
feat: [PL-26212]: Use Harness FF client SDK for feature flags

knagurski:
This PR integrates the Harness Feature Flags React client SDK into Harness Core UI and provides env vars to control its configuration.

If the `USE_LEGACY_FEATURE_FLAGS` env variable is set as `true` (or left undefined), the legacy API endpoint will be used. As the new SDK requires an SDK key be defined, the legacy method will remain the default until keys can be created and distributed. Setting `USE_LEGACY_FEATURE_FLAGS` as `false` and providing a `HARNESS_FF_SDK_KEY` value will trigger the loading of the SDK.

 | Variable                       | Default value                                         | Description                                                                                    |
 |--------------------------------|-------------------------------------------------------|------------------------------------------------------------------------------------------------|
 | `HARNESS_FF_SDK_KEY`           | _n/a_                                                 | SDK key to use with the Harness FF client SDK                                                  |
 | `HARNESS_FF_SDK_ENABLE_STREAM` | `false`                                               | Whether or not to enable real-time feature flag streaming when using the Harness FF client SDK |
 | `HARNESS_FF_SDK_BASE_URL`      | `https://config.feature-flags.uat.harness.io/api/1.0` | Base URL to use with the Harness FF client SDK                                                 |
 | `USE_LEGACY_FEATURE_FLAGS`     | `true`                                                | Whether or not to use the legacy feature flags endpoint over using the Harness FF client SDK   |
 | `HARNESS_FF_SDK_ASYNC`         | `true`                                                | Whether of not to use the SDK's async mode or to wait for the initial load of flags before rendering |


diptiman-adak:
 


            
 #### 23 Issues 
* Code issue 23 Code issues 

 [Details](https://app.embold.io/organization/gh/harness/repositories/e864badd0b95ab81c060b15b85f15701/pull-request/11162)

udayvunnam:
:tada: LGTM! @knagurski how are we planning the transition? Entire App or by MFE or individual page/team on their own

udayvunnam (approved):

willd0yle (dismissed):

knagurski:
> 🎉 LGTM! @knagurski how are we planning the transition? Use it as switch for entire App or by MFE or individual page/team on their own

It'll be by environment. Initially I plan to configure Pre-QA to use the new SDK, evaluate how that works out and then move through the envs. Ideally this will be totally transparent for all teams, though we will need to pass on an SDK key to all devs targeting a specific env (or explain how to generate their own using the UAT FF project). At the moment I've made it so that the legacy API will continue to be used.

Regarding MFEs, if they are querying flags through the usual `useFeatureFlag` or `useFeatureFlags` hooks, they should have no issues at all and everything should continue to work as these are fed using the existing `AppStoreProvider`, which in turn is now fed from the SDK/legacy API. It should be possible to also pass along (or directly import) the SDK's version of `useFeatureFlag`/`useFeatureFlags`/`withFeatureFlags`/`ifFeatureFlag`

src/framework/FeatureFlags/FeatureFlagsProvider.tsx
```diff
@@ -0,0 +1,42 @@
...
+    }),
+    [accountId]
+  )
+
+  if (USE_LEGACY_FEATURE_FLAGS) {
+    return <LegacyFeatureFlagsProvider>{children}</LegacyFeatureFlagsProvider>
+  }
+
+  return (
+    <FFContextProvider
```
abhinavrastogi-harness, line 30: does the `FFContextProvider` block on a network call to fetch feature flags?

src/framework/FeatureFlags/FeatureFlagsProvider.tsx
```diff
@@ -0,0 +1,42 @@
...
+  const target = useMemo<FFContextProviderProps['target']>(
+    () => ({
+      identifier: accountId,
+      name: accountId
+    }),
+    [accountId]
+  )
+
+  if (USE_LEGACY_FEATURE_FLAGS) {
+    return <LegacyFeatureFlagsProvider>{children}</LegacyFeatureFlagsProvider>
```
abhinavrastogi-harness, line 26: With this change, the legacy FF api call and rest of startup API calls will become sequential.
They happen in parallel currently.
This can have a drastic negative impact on app load performance.

configs/webpack.common.js
```diff
@@ -185,7 +207,11 @@ const config = {
     ),
     new webpack.ContextReplacementPlugin(/moment[\/\\]locale$/, /en/),
     new webpack.DefinePlugin({
-      'process.env': '{}' // required for @blueprintjs/core
+      'process.env': '{}', // required for @blueprintjs/core
+      HARNESS_FF_SDK_BASE_URL: JSON.stringify(HARNESS_FF_SDK_BASE_URL),
+      HARNESS_FF_SDK_KEY: JSON.stringify(HARNESS_FF_SDK_KEY),
+      HARNESS_FF_SDK_ENABLE_STREAM,
+      USE_LEGACY_FEATURE_FLAGS
```
abhinavrastogi-harness, line 214: i think these need to be runtime injections instead of build-time injections.
We might want different baseurl and keys for the same build in different environments (eg. QA and Prod).

src/framework/FeatureFlags/LegacyFeatureFlagsProvider.tsx
```diff
@@ -0,0 +1,30 @@
...
+  const { data: featureFlags, loading } = useGetFeatureFlags({ accountId })
+
+  const flags = useMemo(
+    () =>
+      fromPairs(
+        featureFlags?.resource?.map(flag => {
+          return [flag.name, !!flag.enabled]
+        })
+      ),
+    [featureFlags?.resource]
```
abhinavrastogi-harness, line 26: this will memoize on the reference to the `resource` key. are we sure this will regenerate the flags when we get a new response?
`featureFlags` is just a state variable.. so I am not sure if the reference will change.

abhinavrastogi-harness (changes requested):
Looks good overall. Few key things that I might consider as blockers:

1. support for reading baseurl, key, legacy flag etc from the runtime env. this will require injection of these values via `.Dockerfile` env and `index.html`
2. making feature-flags api call sequential is a performance regression. do we have any alternatives?

configs/webpack.common.js
```diff
@@ -36,6 +36,28 @@ const enableSCM = process.env.ENABLE_SCM === 'true'
...
+
+console.log('\nFeature flags env vars')
+console.table({ HARNESS_FF_SDK_KEY, HARNESS_FF_SDK_ENABLE_STREAM, HARNESS_FF_SDK_BASE_URL, USE_LEGACY_FEATURE_FLAGS })
+
+if (!USE_LEGACY_FEATURE_FLAGS) {
+  if (!HARNESS_FF_SDK_KEY) {
+    throw new Error('A Harness Feature Flags SDK key must be provided if `USE_LEGACY_FEATURE_FLAGS` is disabled')
+  }
+
+  if (!HARNESS_FF_SDK_BASE_URL.match(/https?:\/\//)) {
```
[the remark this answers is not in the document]
knagurski: Good idea. I just added it as a _crossing t's, dotting i's_ kinda thing, but it's worth doing right!

src/framework/FeatureFlags/FeatureFlagsProvider.tsx
knagurski: Yes, it will display the passed `fallback` component until the initial load is complete, then it will render the children.

src/framework/FeatureFlags/FeatureFlagsProvider.tsx
knagurski: Yes, that's the case. I think I've figured out a pretty simple fix for this, but in my testing, as-is, it adds just under 1s to page load. Not too drastic, but I think we can do better... instead of having a `LegacyFeatureFlags` provider, we can instead have `AppStoreContext` make the API call only when `withFeatureFlags` passes an empty result set. So when the SDK is configured and fetches, it uses those flags and when it is bypassed (with `USE_LEGACY_FEATURE_FLAGS`), it just does what it does currently.

configs/webpack.common.js
knagurski: That's a great call. Do we have an example of this being done for other features?

src/framework/FeatureFlags/LegacyFeatureFlagsProvider.tsx
knagurski: So, you're both correct and ¯\_(ツ)_/¯ at the same time. `resource` doesn't exist when the memo is first called, but does when the API call completes. Since the API will never be called again, I'm treating `resource` as a boolean to trigger the memo; it's contents will never change. In the FF React SDK, I do a similar thing, but use `JSON.stringify` on the result to change on key/value changes in the variable; but that's overkill in this case where I just want to know it exists.

vkbansal-harness (approved):

udayvunnam (approved):

package.json
```diff
@@ -64,6 +64,7 @@
     "@harness/telemetry": "^1.0.42",
     "@harness/uicore": "3.80.0",
     "@harness/use-modal": "1.1.0",
+    "@harnessio/ff-react-client-sdk": "^1.0.0-alpha.10",
```
abhinavrastogi-harness, line 67: Can we mvoe to a stable version instead of 1.0.0-alpha.10 for the sdk?

src/index.html
```diff
@@ -140,6 +140,28 @@
...
     <!-- End of Saber Feedback button -->
+
+    <!-- Harness Feature Flags -->
+    <script>
+      const config = {
+        useLegacyFeatureFlags: 'USE_LEGACY_FEATURE_FLAGS' !== 'false',
+        baseUrl: 'HARNESS_FF_SDK_BASE_URL' || 'https://config.feature-flags.uat.harness.io/api/1.0',
+        enableStream: 'HARNESS_FF_SDK_ENABLE_STREAM' === 'true',
+        sdkKey: 'HARNESS_FF_SDK_KEY',
+        async: 'HARNESS_FF_SDK_ASYNC' !== 'false'
```
abhinavrastogi-harness, line 151: Two potential improvements purely for semantics:
1. we usually append `_PLACEHOLDER` to anything that is meant to be replaced as a template variable.. eg. `HARNESS_FF_SDK_KEY_PLACEHOLDER` instead of `HARNESS_FF_SDK_KEY`
2. when this is merged, if there is any environment where the correct service variables/overrides are not present, something like 
```
'HARNESS_FF_SDK_BASE_URL' || 'https://config.feature-flags.uat.harness.io/api/1.0'
```
will always evaluate to `HARNESS_FF_SDK_BASE_URL` as a string instead of undefined & using the prod URL.

To solve this, we can use HTML comments to set the default values.. similar to other variables on line 38.

yarn.lock
```diff
@@ -11485,6 +11506,11 @@ lodash.merge@^4.6.2:
   resolved "https://registry.yarnpkg.com/lodash.merge/-/lodash.merge-4.6.2.tgz#558aa53b43b661e1925a0afdfa36a9a1085fe57a"
   integrity sha512-0KpjqXRVvrYyCsX1swR/XTK0va6VQkQM6MNo7PqW77ByjAhoARA8EfrP1N4+KlKj8YS0ZUCtRT/YUuhyYDujIQ==
 
+lodash.omit@^4.5.0:
```
abhinavrastogi-harness, line 11509: does the SDK bring this dependency? 
Why are these not included in the SDK bundle?

package.json
knagurski: Not yet. The SDK was written specifically for this purpose. The code is complete and tested, but the docs still need written and the SDK publicised. When those things are in place, we'll go GA. The `alpha` part more refers to it not being ready for public consumption, not an indicator of its stability.

src/index.html
knagurski: Regarding _"will always evaluate to HARNESS_FF_SDK_BASE_URL as a string instead of undefined & using the prod URL"_ because it will always be set by the docker script, it will either be a value (when the source variable is defined), or it will be blank. In the case of a blank string, it will be evaluated as falsey and fallback to the API.

knagurski: Regarding _"we usually append `_PLACEHOLDER` to anything that is meant to be replaced as a template variable"_ I saw an even mix of things using `_PLACEHOLDER` and not, so I just went with an example that was closer to how I was doing things. It's a good idea though, will update now 👍🏻

yarn.lock
knagurski: Bundling things with a library is a pretty wasteful practice and is specifically what `dependencies` are for. By bundling it we take away the ability for npm/yarn to handle version negotiation, meaning the final app's bundle size will definitely increase, rather than only _potentially_ increasing as happens with a `dependency`. IMO, bundling is something that should _only_ be done by the final app, not any of the upstream dependencies.

src/index.html
knagurski: Updated now with some local dev goodness and `_PLACEHOLDER` suffix

yarn.lock
knagurski: I should actually be able to eliminate this dependency completely in an upcoming release of the SDK

abhinavrastogi-harness (approved):

package.json
knagurski: UPDATE: I'm reviewing a draft for the docs, so we're pretty close to a GA on the SDK 👍🏻 No code changes should be involved between `1.0.0-alpha.12` and `1.0.0`

willd0yle (approved):

PR #11908, opened 2022-10-17
fix: [PL-28843]: Avoiding unnecessary git experience related API calls for pipeline Studio

sunnykesh-harness:
### Summary

Avoiding total of 4 API calls when user navigated to pipelineStudio by a shared link or reload the page.

RCA of issue and explanation for changes : 

1. It was observed `isGitSyncEnabled` API was called 3 times. Although it is not impacting application performance right now as there are other [performance bottlenecks listed here.](https://harness.atlassian.net/wiki/spaces/~5c57a22c0f8313373e56a974/pages/21180252475/NGUI+Performance+Analysis+13+Sep+2022)  but once those fixed this will be new performance bottleneck
   **Note :** Other than reload or loading Pipeline Studio directly, while using the application this API was getting called only once as expected when selected project was changed.
 In App store with  isGitSyncEnabled response + few FFs decide whether we have to support oldGitSync or new gitExperience, and if we have to support oldGitSync then it has been enabled for the selected project or not.

 Reasons for API called 3 times:

- 1st time it was called as default after application load.
- Just after application load, appStore was setting selected project from URL and we call isGitSyncEnabled whenever selectedProject changes.
- 3rd time it was called after FF API response.

   **Fix :** 
   Added the check to call the API when FF API response is available.

2. Conditionally wrapping PageHeader in GitSyncStoreProvider(calling list of oldGitSyncRepos, list of SourceCodeManagers) only for oldGitSync pipelines. This will avoid 2 API calls for INLINE and new gitExperience REMOTE pipelines.
Note: GitSyncStoreProvider has many other usage which should be conditionally or avoided completely.

#### Screenshots

NA

#### PR Checklist

<details>

</details>

<details>

- Jest: `retrigger jest`
- ESLint: `retrigger eslint`
- Build: `retrigger build`
- Rebase: `trigger rebase`
- Cypress: `trigger cypress`
</details>


chetanvekariya-harness (approved):

sumanmabd (approved):

depanshu-dhiman (approved):

src/modules/70-pipeline/pages/pipeline-details/PipelineDetails.tsx
```diff
@@ -229,117 +229,121 @@ export default function PipelineDetails({ children }: React.PropsWithChildren<un
     }
   }
 
+  const renderPageHeader = (): JSX.Element => {
```
abhinavrastogi-harness, line 232: can this be a pure component instead of an inline function? it'll be more performant that way.

abhinavrastogi-harness (approved):

depanshu-dhiman (approved):

PR #12059, opened 2022-10-25
feat: [FFM-3176]: Add Flag Changes sub-section validation

knagurski:
This PR adds Yup validation to each of the Flag Changes sub-sections and applies them to the widget and runtime input views of the Feature Flags pipeline step.

|Before|After|
|------|------|
|Before|After|


src/modules/75-cf/components/PipelineSteps/FlagConfigurationStep/FlagChanges/subSections/ServePercentageRollout.tsx
```diff
@@ -6,18 +6,48 @@
...
+            Yup.object({
+              values: Yup.array()
+                .default([''])
+                .of(
+                  Yup.string()
+                    .trim()
+                    .min(1, getString('cf.featureFlags.flagPipeline.validation.servePercentageRollout.targetGroup'))
+                )
+            })
+          ),
```
knagurski, line 36: All of this is to catch `clauses[0].values[0]` to match the expected instruction's Target Group id.

laura-mannus (approved):
Nice work!

Kira-Jennings (approved):

PR #12147, opened 2022-10-31
feat: [PIE-6094]: allowed values for custom number variables

AshwinBhatkal-Harness:
### Summary

Allowed values for custom number variables

#### Screenshots

https://user-images.githubusercontent.com/96711906/199078116-25aef199-6d92-4039-b7c5-c0cdd508f784.mov

#### PR Checklist

<details>

</details>

<details>

- Jest: `retrigger jest`
- ESLint: `retrigger eslint`
- Build: `retrigger build`
- Rebase: `trigger rebase`
</details>


src/modules/70-pipeline/components/PipelineSteps/Steps/CustomVariables/CustomVariableInputSet.tsx
```diff
@@ -159,34 +155,20 @@ function CustomVariableInputSetBasic(props: ConectedCustomVariableInputSetProps)
                 />
               ) : (
                 <>
-                  {parsedInput?.allowedValues?.values ? (
```
AshwinBhatkal-Harness, line 162: This condition when evaluated to true handled array of default values while handling a single default value when evaluated to false

src/modules/10-common/components/ConfigureOptions/ConfigureOptionsUtils.ts
```diff
@@ -200,7 +200,8 @@ export function parseInput(input: string): ParsedInput | null {
       const jexlMatch = fnArgs.match(JEXL_REGEXP)
 
       parsedInput[InpuSetFunction.ALLOWED_VALUES] = {
-        values: jexlMatch ? null : fnArgs.split(','),
+        // map is required to retain original type of value
+        values: jexlMatch ? null : fnArgs.split(',').map(fnArg => yamlParse(fnArg)),
```
AshwinBhatkal-Harness, line 204: Currently yaml parse is strictly defined with type, so I either have to declare it as so - yamlParse<string>(fnArg) or just yamlParse(fnArg) as fnArg is of type string

If you know how I can use .map(yamlParse) without typescript throwing an error, it would be cool.

ishantmehtaharness: Is yamlParse a cheap operation or an expensive one?
We are explicitly doing yamlParse for all scenarios 
Do we need this for all scenarios or for a limited set of scenarios?

ishantmehtaharness, line 204: Is yamlParse a cheap operation or an expensive one?
We are explicitly doing yamlParse for all scenarios
Do we need this for all scenarios or for a limited set of scenarios?

ishantmehtaharness (commented):
Added comment

src/modules/10-common/components/ConfigureOptions/ConfigureOptionsUtils.ts
AshwinBhatkal-Harness: YamlParse is from js-yaml. It is very fast. I'm not sure if it is a cheap operation though. I'm guessing since it is fast, they have optimised it with the input given.

Also, if we don't do it for all scenarios, then we may have to add specific handling for different types. We may have to loop through an array, check the type, check it's actual type and then parse. This is exactly what yamlParse already does

AshwinBhatkal-Harness: YamlParse is from js-yaml. It is very fast. I'm not sure if it is a cheap operation though. I'm guessing since it is fast, they have optimised it with the input given.

Also, if we don't do it for all scenarios, then we may have to add specific handling for different types. We may have to loop through an array, check the type, check it's actual type and then parse. This is exactly what yamlParse already does

ishantmehtaharness: Sure
In that case we can merge it

ishantmehtaharness (approved):
Looks good

PR #12256, opened 2022-11-07
fix: [FFM-5203]: Ensure correct instructions are displayed when onboarding

knagurski:
This PR fixes an issue whereby the only the Xamarin instructions could be shown during onboarding. It also fixes some issues with misspellings and broken PHP code in strings file.

|Before|After|
|------|------|
|Before|After|


crystal-fong (dismissed):

src/modules/75-cf/pages/onboarding/views/SetUpYourApplicationView.tsx
```diff
@@ -43,6 +43,12 @@ export const SetUpYourApplicationView: React.FC<SetUpYourApplicationViewProps> =
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [])
 
+  useEffect(() => {
+    if (language?.name !== 'Xamarin') {
```
chrisblakely01, line 47: why do we have to be specific about Xamarin in this file?

knagurski: Xamarin is the only one that has language-specific branching. Xamarin isn't so much a language as it is a cross-platform library, so we branch for iOS and Android.

knagurski: The reason we have this `useEffect` is that `currentReadme` is actually an override for the readme that's referenced in the `language`. Xamarin is (currently) the only "language" that we need to override for. Without this `useEffect`, selecting _Xamarin_ and then switching to another language, the override would still be in place and we'd show the wrong thing. Another approach would be to do this check when rendering the `SetUpYourCodeView` component on line 129, but this felt like the least intrusive option for this bug fix.

src/modules/75-cf/strings/strings.en.yaml
```diff
@@ -1545,7 +1545,7 @@ onboarding:
       use Harness\Client;
       use OpenAPI\Client\Model\Target;
 
-      $SDK_KEY = getenv({{apiKey}}) ?: "";  // you can put your key in env variable or you can provide in the code
+      $SDK_KEY = getenv("SDK_KEY") ?: "";  // you can put your key in env variable or you can provide in the code
```
crystal-fong, line 1548: I think `{{apiKey}}` grabs the value of the generated API key so users don't need to manual copy/paste it in themselves

src/modules/75-cf/pages/onboarding/views/SetUpYourApplicationView.tsx
laura-mannus: Nice

src/modules/75-cf/strings/strings.en.yaml
knagurski: In this case it would produce broken PHP code. The `getenv` function gets an env var by name, so this would plop the actual key down as a name of an env var to get, which won't work. Worse, it'll plop it down not in a string, so it'll try to reference the actual key as a constant (not even a variable, since there's no `$`). Also that key will have dashes, so this will then evaluate as being a bunch of non-initialised constants trying to be subtracted from each other. In other words, it'll break the example pretty completely :-p

knagurski: What you'd probably want is just

```php
$SDK_KEY = "{{ apiKey }}";
```

crystal-fong: ^^ why I don't deal with php

chrisblakely01 (approved):

crystal-fong (approved):

src/modules/75-cf/strings/strings.en.yaml
knagurski: To be fair JS would do the exact same thing (assuming there was a `getenv` function)

laura-mannus (approved):

PR #12307, opened 2022-11-09
feat: [SPG-871]: remove external fonts for SMP env

abhinavrastogi-harness:
### Summary

Avoid loading Google Fonts in SMP environment.

Adding the font link conditionally in the HTML directly from docker for SaaS. This ensures fonts are loaded without waiting for script execution in prod env.

Adding the font link via JS for local dev.

#### Screenshots

NA

#### PR Checklist

<details>

</details>

<details>

- Jest: `retrigger jest`
- ESLint: `retrigger eslint`
- Build: `retrigger build`
- Rebase: `trigger rebase`
</details>


vkbansal-harness (approved):

udayvunnam (approved):

knagurski:
Instead of loading logic and using the Google Web Fonts APIs, we can just take the actual font files we use and embed them in the project itself. All Google Font typefaces allow local serving as part of the license. We then won't have the 3rd party calls or need to worry about what env it's in.

src/index.html
```diff
@@ -79,11 +78,18 @@
       window.newNavContentfulAccessToken = '<%= NEW_NAV_CONTENTFUL_ACCESS_TOKEN_DEV %>'
       window.newNavContetfulSpace = '<%= NEW_NAV_CONTENTFUL_SPACE_DEV %>'
       window.newNavContentfulEnvironment = '<%= NEW_NAV_CONTENTFUL_ENVIRONMENT_DEV %>'
+
+      function createFontLink(href) {
```
knagurski, line 82: I may be reading this wrong, but won't this only run in dev mode?

abhinavrastogi-harness:
I am looking into embedding the font files in the project itself, but running into some challenges with it. We can take it up separately.
But in either case, I am only planning to load the local font files only for SMP.
Using Google Fonts in SaaS gives us the benefit of popular fonts already being cached on the user's browser because they might have visited another website which uses the same font.

knagurski:
> I am looking into embedding the font files in the project itself, but running into some challenges with it. We can take it up separately.
> But in either case, I am only planning to load the local font files only for SMP.
> Using Google Fonts in SaaS gives us the benefit of popular fonts already being cached on the user's browser because they might have visited another website which uses the same font.

No problem. Regarding the point about using Google Fonts as they might be already cached; that was pointed out as being a fallacy by the Google engineers. In their testing, cache hits from the Google Fonts endpoints were exceedingly rare (browsers will purge this style of cache far more often to make way for website-specific caches) and (again in their tests) the establishing of a connection to the 3rd party to check the etag is still fresh was substantially slower and less reliable than serving from the domain with valid caching strategies. Was a great talk actually that went super deep on really esoteric performance stuff. I'll see if I can dig it out, but their recommendation was to only use the Google Fonts service itself when prototyping and never for production.

src/index.html
```diff
@@ -89,11 +89,17 @@
...
+        e.setAttribute('href', href)
+        e.setAttribute('rel', 'stylesheet')
+        document.getElementsByTagName('head')[0].appendChild(e)
+      }
+      createFontLink('https://fonts.googleapis.com/css2?family=Inter:wght@400;500;600&display=swap')
+      createFontLink('https://fonts.googleapis.com/css2?family=Roboto+Mono:wght@300;700&display=swap')
+      createFontLink('https://fonts.googleapis.com/css2?family=Reenie+Beanie&display=swap')
     </script>
-    <% } %>
-    <!-- Enabling ng-auth-ui -->
```
vkbansal-harness, line 94: these comments removed intentionally?

src/index.html
```diff
@@ -89,11 +89,17 @@
...
+        var e = document.createElement('link')
+        e.setAttribute('href', href)
+        e.setAttribute('rel', 'stylesheet')
+        document.getElementsByTagName('head')[0].appendChild(e)
+      }
+      createFontLink('https://fonts.googleapis.com/css2?family=Inter:wght@400;500;600&display=swap')
+      createFontLink('https://fonts.googleapis.com/css2?family=Roboto+Mono:wght@300;700&display=swap')
+      createFontLink('https://fonts.googleapis.com/css2?family=Reenie+Beanie&display=swap')
     </script>
-    <% } %>
```
vkbansal-harness, line 93: You can insert link tags directly here. This block is for dev mode.

udayvunnam (dismissed):

src/index.html
abhinavrastogi-harness: yes

src/index.html
abhinavrastogi-harness: moved JS logic to webpack dev flag in HEAD tag itself

vkbansal-harness (approved):

udayvunnam (approved):

PR #12582, opened 2022-11-22
fix: [FFM-5086]: fix having to scroll for Back/Next buttons when content overflows container

Kira-Jennings:
### Summary
- fix having to scroll for Back/Next buttons when content overflows container
- also fixed stepper to not be hidden behind MarkdownViewer readme content

#### Screenshots
Pasted Graphic 12

#### PR Checklist

<details>

</details>

<details>

- Jest: `retrigger jest`
- ESLint: `retrigger eslint`
- Build: `retrigger build`
- Rebase: `trigger rebase`
</details>


Kira-Jennings:
retrigger titlecheck

laura-mannus:
Hey, I know you added this in 
image
was this a requirement from Ksenia?

laura-mannus:
You can scroll in the last page when you don't need to 
image
not a biggie just noticed it

Kira-Jennings:
> NP: You can scroll in the last page when you don't need to
Good catch!

Kira-Jennings:
> Hey, I know you added this in (that the stepper now scrolls with you) was this a requirement from Ksenia?
@laura-mannus I didn't do this deliberately, so I'll check with her if she wants it to behave like that

laura-mannus (approved):
LGTM

src/modules/75-cf/pages/onboarding/OnboardingDetailPage.tsx
```diff
@@ -129,7 +134,7 @@ export const OnboardingDetailPage: React.FC = () => {
           }
         />
       </Layout.Horizontal>
-      <Container style={{ flexGrow: 1 }} padding={{ top: 'xxlarge', bottom: 'xxlarge' }}>
+      <Container className={css.scrollOverflowY} height="100vh" padding={{ top: 'xxlarge', bottom: 'xxlarge' }}>
```
chrisblakely01, line 137: Not sure if this applies - but we're not meant to use `100vh` for page stuff anymore (the sidenav has so much crap in it now it typically overflows, causing the page content to crop)

theres a var to use instead 

```
var(--min-page-height);
```

chrisblakely01 (approved):
LGTM - just one thing that might not be a thing

src/modules/75-cf/pages/onboarding/OnboardingDetailPage.tsx
knagurski, line 137: With this being `100vh` and the previous sibling being 10px, won't this always overflow, regardless of the content?

knagurski (approved):
Good stuff 👍🏻

src/modules/75-cf/pages/onboarding/OnboardingDetailPage.tsx
Kira-Jennings: Grid will be added in the next ticket https://harness.atlassian.net/browse/FFM-5084

Kira-Jennings: @knagurski  The height isn't being ignored, without setting it the Back/Next buttons end up not being positioned at the bottom of the page - instead they're right below the end of the content
image

Kira-Jennings: I didn't know about this, thanks for the heads up

knagurski: It's not that it's being ignored, it's that it's not being honoured. That panel is not `100vh`, it's more like `80vh`. What's happening here is that the parent is set as `100vh` and to use flexbox, since the browser wants to keep it in the `100vh` of the parent and this panel allows overflow, it shrinks this panel to fit. You can confirm this by changing the panel's `100vh` to something stupid big (like `200vh`). It'll just grow to fill the available space, or in this case, shrink to fit.
Screenshot 2022-11-24 at 10 56 41

PR #12613, opened 2022-11-23
fix: [FFM-4582]: Restyle flag pipeline card

knagurski:
This PR changes the styling of the Flag Pipelines card as well as fixing a bug whereby the user could get stuck if they searched for a pipeline and received no results.

|Before|After|
|------|------|
|Before|After|


src/modules/75-cf/pages/feature-flags-detail/flag-pipeline-tab/components/available-pipelines-drawer/AvailablePipelinesDrawer.module.scss
```diff
@@ -27,9 +28,8 @@
 
 .pipeline-cards-grid {
   display: grid;
-  grid-template-columns: 1fr 1fr 1fr;
-  grid-auto-rows: 250px;
-  gap: var(--spacing-xxlarge);
+  grid-template-columns: repeat(auto-fill, 198px);
```
knagurski, line 31: This 198px really annoys me. I played around with using something like `minmax(150px, 1fr)`, but it always grows to the `1fr` and means the 3 column layout turns to 2 and when there's only a single pipeline it stretches. So I gave up and locked it. Unfortunately `200px` was 2px too much to fit in a 3 col layout. I also considered using `repeat(3, 1fr)`, but when this is on a smaller screen, I do want to allow 2 and 1 col as spare dictates.

src/modules/75-cf/pages/feature-flags-detail/flag-pipeline-tab/components/available-pipelines-drawer/AvailablePipelinesDrawer.tsx
```diff
@@ -149,10 +149,10 @@ const AvailablePipelinesDrawer: React.FC<AvailablePipelinesDrawerProps> = ({
             aria-label={getString('close')}
           />
         </Container>
-        <Text font={{ variation: FontVariation.BODY2 }} color={Color.GREY_500} tag="div">
+        <Text font={{ variation: FontVariation.BODY2 }} color={Color.GREY_500} tag="div" className={css.description}>
           <String stringID="cf.featureFlags.flagPipeline.drawerDescription" useRichText />
         </Text>
-        {pipelinesAvailable && (
+        {(pipelinesAvailable || searchTerm) && (
```
knagurski, line 155: Without this extra check, if you search for a pipeline and get no results, there is no way to clear the search. Moving it to state allowed me to kill off the use of the `refetch` function and have the fetch driven off the state instead.

src/modules/75-cf/pages/feature-flags-detail/flag-pipeline-tab/components/available-pipelines-drawer/AvailablePipelinesDrawer.tsx
```diff
@@ -170,8 +170,10 @@ const AvailablePipelinesDrawer: React.FC<AvailablePipelinesDrawerProps> = ({
                 {availableFeaturePipelines?.availablePipelines.map(pipeline => (
                   <PipelineCard
                     key={pipeline.identifier}
+                    identifier={pipeline.identifier}
                     pipelineName={pipeline.name}
                     pipelineDescription={pipeline.description}
+                    lastUpdatedAt={pipeline.lastUpdatedAt as number}
```
knagurski, line 176: The typings on `lastUpdatedAt` seem to be wrong as it is always present and starts off being the same as the creation time.

src/modules/75-cf/pages/feature-flags-detail/flag-pipeline-tab/components/pipeline-card/PipelineCard.module.scss
```diff
@@ -5,10 +5,28 @@
...
+    gap: var(--spacing-xsmall);
+  }
+
+  footer {
+    border-top: 1px solid var(--grey-100);
+  }
+}
+
+.icon {
+  transform: translateX(-2px);
```
knagurski, line 30: The FF icon has 2px of whitespace at this size, so we translate it to the left by that much to optically bring it into alignment

src/modules/75-cf/pages/feature-flags-detail/flag-pipeline-tab/components/pipeline-card/PipelineCard.tsx
```diff
@@ -6,30 +6,58 @@
...
+
   return (
-    <Card className={isSelected ? cx(css.cardActive, css.card) : css.card} onClick={onClick} role="listitem">
-      <Layout.Vertical spacing="small">
-        <Icon name="ff-solid" size={30} />
-        <Text font={{ variation: FontVariation.BODY1 }}>{pipelineName}</Text>
-        {pipelineDescription && (
-          <Text font={{ variation: FontVariation.BODY2 }} color={Color.GREY_500}>
-            {pipelineDescription}
+    <Card selected={isSelected} onClick={onClick} role="listitem" className={css.card}>
```
knagurski, line 36: I removed non-standard styling for the selected state to bring it inline with the rest of Harness.

devinder-harness (approved):

src/modules/75-cf/pages/feature-flags-detail/flag-pipeline-tab/components/available-pipelines-drawer/AvailablePipelinesDrawer.tsx
```diff
@@ -170,8 +170,10 @@ const AvailablePipelinesDrawer: React.FC<AvailablePipelinesDrawerProps> = ({
                 {availableFeaturePipelines?.availablePipelines.map(pipeline => (
                   <PipelineCard
                     key={pipeline.identifier}
+                    identifier={pipeline.identifier}
```
abhinavrastogi-harness, line 173: it could be cleaner to just pass the entire `pipeline` object to the PipelineCard component.

abhinavrastogi-harness (approved):

src/modules/75-cf/pages/feature-flags-detail/flag-pipeline-tab/components/available-pipelines-drawer/AvailablePipelinesDrawer.tsx
knagurski: I know... I was thinking the same thing. I do like the idea of keeping the knowledge of the child component to a minimum to reduce coupling, but at this point most of the original object is being used, so just passing it wholesale does seem cleaner. Next time I'm in the area, I think I'll do a refactor 👍🏻

Kira-Jennings (approved):

PR #12639, opened 2022-11-24
fix: [FFM-5507]: Simplify flag listing logic to eliminate double API calls

knagurski:
This PR removes unnecessary API calls by simplifying the logic used to trigger a fetch. It also removes some code that tries to optimistically update the UI after a flag is toggled, but in practices didn't trigger a rerender and so didn't really do anything.


src/modules/75-cf/pages/feature-flags/FeatureFlagsPage.tsx
```diff
@@ -467,32 +471,13 @@ const FeatureFlagsPage: React.FC = () => {
...
-    () => toggleFeatureFlag.loading || deleteFlag.loading,
-    [toggleFeatureFlag.loading, deleteFlag.loading]
-  )
 
   const gitSync = useFFGitSyncContext()
-
   const governance = useGovernance()
 
+  const loading = flagsLoading || envsLoading
+  const gitSyncing = toggleFeatureFlag.loading || deleteFlag.loading
```
knagurski, line 480: These were initially `useMemo` calls, but their actual logic takes less time to execute than the memoization and look up.

src/modules/75-cf/pages/feature-flags/FeatureFlagsPage.tsx
```diff
@@ -509,18 +494,6 @@ const FeatureFlagsPage: React.FC = () => {
...
-                if (feature) {
-                  if (feature.envProperties) {
-                    feature.envProperties.state = (
-                      status ? FeatureFlagActivationStatus.ON : FeatureFlagActivationStatus.OFF
-                    ) as FeatureState
-                  }
-                  feature.modifiedAt = Date.now()
-                  setFeatures({ ...features } as Features)
-                }
-              }}
```
knagurski, line 523: This looks like it's trying to do an optimistic update of the UI whilst the API call completes, but in practice it doesn't trigger a rerender, so doesn't actually change anything in the UI.

src/modules/75-cf/pages/feature-flags/FeatureFlagsPage.tsx
```diff
@@ -149,9 +145,8 @@ export const RenderColumnFlag: React.FC<RenderColumnFlagProps> = ({
       }
 
       setStatus(!status)
-      update(!status)
       refetchFlags()
-    } catch (error: any) {
+    } catch (error) {
```
knagurski, line 149: TS already knows what this error is, so can drop the `:any`

crystal-fong (approved):

Kira-Jennings (approved):

laura-mannus (approved):

PR #13117, opened 2022-12-20
fix: [FFM-5187]: use portal for Target Group modal dropdowns

knagurski:
This PR enables the `usePortal` option for the dropdowns in the _Edit Target Group Criteria_ modal to prevent unnecessary scrolling.

|Before|After|
|------|------|
|Before|After|


Harness-JKhattak (approved):
Should we be aiming to switch other modules over to use portals that use these Select inputs?

knagurski:
> Should we be aiming to switch other modules over to use portals that use these Select inputs?

Maybe. TBH the drop downs are kinda poor at the moment and work needs done on them, but by using a portal, it does remove one of their biggest issues. I'd treat it case-by-case.

pankajsingh-harness (approved):

PR #13122, opened 2022-12-20
fix: [CDS-48598]: Update Styles for GitOps Entities in Account Resources

ishantmehtaharness:
Screenshot 2022-12-20 at 10 44 34 PM


src/modules/10-common/strings/strings.en.yaml
```diff
@@ -5,7 +5,7 @@ help: Help
 dashboards: Dashboards
 setup: Setup
 agent: Agent
-agents: Agents
+agents: Gitops Agents
```
rameshx, line 8: This changes the text everywhere `agents` is used. Intentional?
Is it better to rename the key too?

rameshx (approved):

src/modules/10-common/strings/strings.en.yaml
pankajsingh-harness: This has been used at https://github.com/harness/harness-core-ui/blob/develop/src/modules/75-cd/RouteDestinations.tsx#L100 along with the `common.agent`. Please verify if both need to be updated.

src/modules/10-common/components/ResourceCardList/ResourceCardList.module.scss
```diff
@@ -41,6 +41,13 @@
   background-color: var(--magenta-100);
 }
 
+.gitOpsIcon {
+  svg path {
+    // We have blue icons today, remove this once we add non-colored icons
```
udayvunnam, line 46: nit: keep `TODO:` prefix to all such things which we can easily find and update later

udayvunnam (approved):

pankajsingh-harness (approved):

udayvunnam (approved):

PR #13170, opened 2022-12-22
fix: [FFM-4385]: Prevent Metrics tab overflow

knagurski:
This PR fixes an issue whereby the metrics tab would double overflow when there are no metrics to display. It also replaces a non-standard loader and no data view.

|Before|After|
|------|------|
|Before|After|


src/modules/75-cf/components/NoData/__tests__/NoData.test.tsx
```diff
@@ -69,4 +69,26 @@ describe('NoData', () => {
...
+
+    expect(document.querySelector('[data-icon]')).not.toBeInTheDocument()
+  })
+
+  test('it should apply icon props', async () => {
+    const size = 200
+    renderComponent({ icon: 'cf-main', imageURL: undefined, iconProps: { size } })
+
+    const svg = document.querySelector(`[data-icon] > svg`)
+    expect(svg).toHaveAttribute('height', `${size}`)
```
knagurski, line 91: These are annoying, but attributes are _always_ strings, which means I needed to cast `size` as a string for the match.

src/modules/75-cf/components/NoData/NoData.tsx
```diff
@@ -7,12 +7,15 @@
...
 import { Button, ButtonProps, ButtonVariation, Container, Heading, Layout, Text } from '@harness/uicore'
+import { Icon, IconName, IconProps } from '@harness/icons'
 import type { LayoutProps } from '@harness/uicore/dist/layouts/Layout'
 import { Color, FontVariation } from '@harness/design-system'
 import css from './NoData.module.scss'
 
 export interface NoDataProps extends LayoutProps {
   imageURL?: string
+  icon?: IconName
+  iconProps?: Partial<IconProps>
```
pankajsingh-harness, line 18: `IconProps` has `name` as `IconName` so do we need to add icon separately.  We can have just `iconProps?: IconProps`

knagurski: Technically yes, but I'm following the pattern that the `Button` component uses in UICore. The `icon` prop makes it simple to add an icon with defaults. The `iconProps` prop gives you extra power, but won't be used as often. The 99% use-case is just `icon="MyIcon"`, which is much more obvious than `iconProps={{ name: 'MyIcon' }}`.

I'm actually more tempted to remove `iconProps` altogether since we don't need it here at the moment 🤔

chrisblakely01 (approved):

PR #13193, opened 2022-12-26
feat: [PIE-7307]: When there is one Rule in Freeze config , on click …

ishantmehtaharness:
**Requirement**:
When there is one Rule in Freeze config , on click of Continue , it should be persisted
This has been a long pending ask from many people

**Solution**:
From parent, we need to call a child component function, on tab change.
I implemented that using formik Ref.

**Coverage failing**: This is covered in Cypress
Screenshot 2022-12-27 at 3 23 33 PM


Eric761 (approved):

src/modules/72-freeze-windows/components/FreezeWindowStudioBody/FreezeWindowStudioVisualView.tsx
```diff
@@ -29,7 +29,11 @@ export const FreezeWindowStudioVisualView = ({ resources }: { resources: Resourc
   const { updateQueryParams } = useUpdateQueryParams<{ sectionId?: string | null }>()
   const { sectionId } = useQueryParams<{ sectionId?: string | null }>()
   const { isReadOnly, isActiveFreeze } = React.useContext(FreezeWindowContext)
-  const formikRef = React.useRef<FormikProps<FreezeObj>>()
+  const formikRef = React.useRef<
+    FormikProps<{ entity?: Array<Record<string, any>> }> & {
```
chetanvekariya-harness, line 33: Can we create new type for `{ entity?: Array<Record<string, any>> }> & { onTabChange: (formikValues: { entity?: Array<Record<string, any>> }) => void }` and use it everywhere ?

chetanvekariya-harness (approved):

src/modules/72-freeze-windows/components/FreezeWindowStudioBody/FreezeWindowStudioVisualView.tsx
ishantmehtaharness: Yes @chetanvekariya-harness , I can do that in a followup PR today.

PR #13273, opened 2023-01-03
fix: [FFM-6221]: Store env in preference store

knagurski:
This PR stores the current env in and retrieves it from the preference store. When accessing the current environment, if it is not in the URL, it will then be looked for in the preference store and then fallback to loading the 1st env from the envs list.

|Before|After|
|------|------|
|Before|After|


src/modules/75-cf/hooks/useEnvironmentSelectV2.tsx
```diff
@@ -59,10 +71,23 @@ export const useEnvironmentSelectV2 = (params: UseEnvironmentSelectV2Params) =>
...
-        setSelectedEnvironment(selectOptions[0])
-        onChange(selectOptions[0], data?.data?.content?.[0], false)
-        rewriteCurrentLocationWithActiveEnvironment(data?.data?.content?.[0].identifier)
+        const preferenceOption = selectOptions.find(({ value }) => value === preferredEnvironment)
+
+        if (preferenceOption) {
+          setSelectedEnvironment(preferenceOption)
+          onChange(
+            preferenceOption,
+            data.data.content.find(({ identifier }) => identifier === preferredEnvironment) as EnvironmentResponseDTO,
```
laura-mannus, line 82:
```suggestion
            data.data.content?.find(({ identifier }) => identifier === preferredEnvironment) as EnvironmentResponseDTO,
```
do we need to do this ^ here?

laura-mannus: Cool thanks for explaining

laura-mannus (approved):

src/modules/75-cf/hooks/useEnvironmentSelectV2.tsx
knagurski: I actually did have it with a bunch of `?`, but then took a second look, realised they weren't needed and TS didn't disagree :-)

willd0yle (approved):

PR #13293, opened 2023-01-04
fix: [FFM-6240]: Configure FF SDK metrics endpoint

knagurski:
This PR ensures that the metrics endpoint is configured by default and can be overridden.


abhinavrastogi-harness (dismissed):

Kira-Jennings (dismissed):

Kira-Jennings (approved):
👍

PR #13297, opened 2023-01-04
feat: [FFM-5932]: Updating the Getting Started flow to Full Screen

laura-mannus:
### Summary

As per PLG request, the 'Getting Started' onboarding flow has now been updated to full-screen behaviour.

#### Screenshots

https://user-images.githubusercontent.com/103498167/210602087-36c5f6ba-8853-424d-8e5f-6393e7b154ec.mov

#### PR Checklist

<details>

</details>

<details>

- Jest: `retrigger jest`
- ESLint: `retrigger eslint`
- Sonar: `retrigger sonar`
- Build: `retrigger build`
- Rebase: `trigger rebase`
</details>


src/modules/75-cf/pages/onboarding/OnboardingDetailPage.module.scss
```diff
@@ -5,15 +5,38 @@
...
+}
+
+.iconContainer {
+  display: flex;
+  justify-content: space-between;
 }
 
 .onboardingContainer {
   overflow-y: scroll;
+  overflow-x: scroll;
```
knagurski, line 21: Since you're specifying both the X and Y as the same thing, you can just use `overflow` instead. Also `scroll` will always show scrollbars, whereas `auto` will only show if they are required.
```suggestion
  overflow: auto;
```

src/modules/75-cf/pages/onboarding/OnboardingDetailPage.tsx
```diff
@@ -111,116 +112,145 @@ export const OnboardingDetailPage: React.FC = () => {
...
+    <>
       <Layout.Horizontal
-        spacing="xsmall"
-        flex
-        data-testid="getStartedProgressStepper"
-        height="10px"
-        padding={{ bottom: 'xlarge' }}
-        className={css.keepOnTop}
+        className={css.iconContainer}
+        height="0"
```
knagurski, line 118: Why `height` of `0`?

src/modules/75-cf/pages/onboarding/OnboardingDetailPage.module.scss
```diff
@@ -5,15 +5,38 @@
...
+.fullScreenContainer {
   overflow: hidden;
-  height: var(--page-min-height);
+  height: calc(var(--page-min-height) - var(--spacing-huge));
+  margin-left: 250px !important;
+}
+
+.iconContainer {
+  display: flex;
+  justify-content: space-between;
```
knagurski, line 16: Since you used a `Layout.Horizontal` it already has a `display` of `flex`. Also you can set the `justify-content` via the `flex` prop.

src/modules/75-cf/pages/onboarding/OnboardingDetailPage.tsx
```diff
@@ -111,116 +112,145 @@ export const OnboardingDetailPage: React.FC = () => {
...
-            apiKey={apiKey as ApiKey}
-            testDone={testDone}
-            setTestDone={setTestDone}
-            environmentIdentifier={selectedEnvironment.identifier}
-          />
-        )}
-      </Container>
-      <Divider />
-      <Layout.Horizontal spacing="small" padding="small">
+        <img src={featureFlagsImg} width="175px" height="50px" />
```
knagurski, line 122: An `alt` prop is required for all `img` tags. If this is purely decorative, then an empty tag is fine. If there is text in the image, it's best to use that as the alt.

src/modules/75-cf/pages/onboarding/OnboardingDetailPage.tsx
```diff
@@ -111,116 +112,145 @@ export const OnboardingDetailPage: React.FC = () => {
...
         <Button
-          text={getString('back')}
-          icon="chevron-left"
+          minimal
```
knagurski, line 124: Instead of setting `minimal` and `withoutBoxShadow`, you might get better results using the `ICON` button variation. You also probably don't need the `width` and `height`, but have a play around and see.
```suggestion
          variation={ButtonVariation.ICON}
```

src/modules/75-cf/pages/onboarding/OnboardingDetailPage.tsx
```diff
@@ -111,116 +112,145 @@ export const OnboardingDetailPage: React.FC = () => {
...
-        />
-        <FlexExpander />
       </Layout.Horizontal>
-    </Layout.Vertical>
+      <Layout.Horizontal className={css.fullScreenContainer} padding={{ left: 'huge' }}>
+        <Layout.Vertical
+          padding={{ top: 'large', right: 'huge', left: 'huge', bottom: 'none' }}
+          background={Color.WHITE}
+          className={css.mainContent}
+          flex
```
knagurski, line 139: Since this is a `Layout.Vertical` it will already be using `display: flex;`, so you can probably drop this.
```suggestion
```

src/modules/75-cf/pages/onboarding/OnboardingDetailPage.tsx
```diff
@@ -111,116 +112,145 @@ export const OnboardingDetailPage: React.FC = () => {
...
-            }
-            onNext()
-          }}
-        />
-        <FlexExpander />
       </Layout.Horizontal>
-    </Layout.Vertical>
+      <Layout.Horizontal className={css.fullScreenContainer} padding={{ left: 'huge' }}>
+        <Layout.Vertical
+          padding={{ top: 'large', right: 'huge', left: 'huge', bottom: 'none' }}
```
knagurski, line 136: You don't need to specify `bottom: 'none'` as that will be the default.
```suggestion
          padding={{ top: 'large', right: 'huge', left: 'huge' }}
```

src/modules/75-cf/pages/onboarding/OnboardingDetailPage.tsx
knagurski, line 124: Can we add an `aria-label` here to let screen readers know what this button does?

src/modules/75-cf/pages/onboarding/OnboardingDetailPage.tsx
```diff
@@ -111,116 +112,145 @@ export const OnboardingDetailPage: React.FC = () => {
...
+          background={Color.WHITE}
+          className={css.mainContent}
+          flex
+        >
+          <Layout.Horizontal
+            spacing="xsmall"
+            flex
```
knagurski, line 143: No need for this here too
```suggestion
```

src/modules/75-cf/pages/onboarding/featureFlags.svg
```diff
@@ -0,0 +1 @@
....1079-1.9668.4082-1.9668 1.5755 0 1.0188.7455 1.5585 1.734 1.5585zm.3171-.8973c-.5094 0-.8737-.2328-.8737-.6815 0-.4689.4082-.6646.9547-.7422.3205-.0438.9614-.1248 1.12-.253v.6106c0 .5769-.4655 1.0661-1.201 1.0661z"/><path d="m94.296 24.4446c1.4 0 2.4324-.641 2.4324-1.9668v-5.2662h-1.2044v.8569h-.0675c-.1855-.3744-.5735-.9243-1.4877-.9243-1.1977 0-2.1389.9378-2.1389 2.6415 0 1.6902.9412 2.5335 2.1355 2.5335.8872 0 1.3022-.4756 1.4911-.8568h.0607v.9817c0 .7624-.506 1.0761-1.201 1.0761-.7354 0-1.0593-.3474-1.2077-.6511l-1.0998.2666c.2227.7253.9682 1.3089 2.2873 1.3089zm.0101-3.0834c-.8063 0-1.2313-.6275-1.2313-1.5823 0-.9412.4183-1.6328 1.2313-1.6328.7861 0 1.2179.6511 1.2179 1.6328 0 .9885-.4386 1.5823-1.2179 1.5823z"/><path d="m102.075 18.5813c-.168-.8771-.87-1.4371-2.0844-1.4371-1.2482 0-2.0984.614-2.095 1.5721-.0034.7556.4622 1.2549 1.4574 1.4607l.884.1856c.476.1046.698.2968.698.5903 0 .3543-.384.6208-.9646.6208-.56 0-.9244-.2429-1.029-.7085l-1.1909.1147c.1519.9514.9514 1.5148 2.2232 1.5148 1.2953 0 2.2093-.6714 2.2133-1.6531-.004-.7388-.479-1.1909-1.458-1.4034l-.8835-.1889c-.5262-.1181-.7354-.3003-.732-.6005-.0034-.3509.3846-.5938.8935-.5938.564 0 .861.307.955.6478z"/></g></svg>
```
knagurski, line 1: We probably don't need this SVG. We already have the FF logo and we can always use real text instead since we also should have the font.

knagurski (commented):
Some really great and concise work here. A test or two would be nice as well.

One thing to consider is the behaviour of the close button. Since we can invoke this stuff from a bunch of places, having the close button always take you to the onboarding page will be a bit unexpected for users. Might be worth seeing if we can find out a way to move back to the page they were on when the flow is invoked.

src/modules/75-cf/pages/onboarding/OnboardingDetailPage.module.scss
laura-mannus: Nice, thank you

src/modules/75-cf/pages/onboarding/OnboardingDetailPage.module.scss
laura-mannus: Awesome, thanks :)

src/modules/75-cf/pages/onboarding/OnboardingDetailPage.tsx
laura-mannus: I set a height of 0 so that the container could fall in line with the progress stepper. So ideally go behind it?

image

is there a better way to do this?

src/modules/75-cf/pages/onboarding/OnboardingDetailPage.tsx
laura-mannus: You have told me this before sorry completely forgot, thank you :)

src/modules/75-cf/pages/onboarding/OnboardingDetailPage.tsx
laura-mannus: This works lovely ^, thank you!

laura-mannus: Of course, done

src/modules/75-cf/pages/onboarding/OnboardingDetailPage.tsx
laura-mannus: Removed, thanks

src/modules/75-cf/pages/onboarding/OnboardingDetailPage.tsx
laura-mannus: Sweet, removed :)

src/modules/75-cf/pages/onboarding/OnboardingDetailPage.tsx
laura-mannus: Removed, thanks

src/modules/75-cf/pages/onboarding/featureFlags.svg
laura-mannus: Ah cool, I will investigate thank you :)

laura-mannus: Kicking myself for not doing this initially

knagurski:
> > Some really great and concise work here. A test or two would be nice as well.
> > One thing to consider is the behaviour of the close button. Since we can invoke this stuff from a bunch of places, having the close button always take you to the onboarding page will be a bit unexpected for users. Might be worth seeing if we can find out a way to move back to the page they were on when the flow is invoked.
> 
> Hey @knagurski so I checked with Ksenia and she says the expected behaviour is: It goes to onboarding first page if nothing has been created. If the flag is created then [x] redirects to FF list page

That is super weird behaviour. If the user is on the (for instance) Targets page, clicks the open this stuff and then says _"Nah"_, they'll 100% expect to be back where they were, not randomly transported to a different part of the site. This manifests to the user as a full-screen modal, so it should act as such. I'd challenge the behaviour with Ksenia as it will be jarring to the user and not act like all other close buttons in the application act.

src/modules/75-cf/pages/onboarding/OnboardingDetailPage.tsx
```diff
@@ -111,116 +113,152 @@ export const OnboardingDetailPage: React.FC = () => {
...
+        <Layout.Horizontal
+          spacing="small"
+          style={{ justifyContent: 'flex-start', alignItems: 'center' }}
+          padding={{ top: 'xxlarge' }}
+        >
+          <Icon data-testid="cfLogo" name="cf-main" height={30} size={30} />
+          <Text font={{ variation: FontVariation.SMALL_BOLD }} color={Color.GREY_800}>
+            {getString('common.purpose.cf.continuous')}
+          </Text>
+        </Layout.Horizontal>
```
knagurski, line 132: I think you should be able to combine the layout, icon and text into a single element if you use the `icon` and `iconProps` props of the `Text` component.
```suggestion
        <Text
          font={{ variation: FontVariation.SMALL_BOLD }}
          color={Color.GREY_800}
          icon="cfLogo"
          iconProps={{ size: 30 }}
        >
          {getString('common.purpose.cf.continuous')}
        </Text>
```

src/modules/75-cf/pages/onboarding/OnboardingDetailPage.tsx
```diff
@@ -111,116 +113,152 @@ export const OnboardingDetailPage: React.FC = () => {
...
         <Button
-          text={getString('back')}
-          icon="chevron-left"
+          icon="code-close"
+          aria-label="close"
```
knagurski, line 135: The label should be treated like any other user-accessible string and come from the strings file instead of being hard coded

src/modules/75-cf/pages/onboarding/OnboardingDetailPage.tsx
```diff
@@ -111,116 +113,152 @@ export const OnboardingDetailPage: React.FC = () => {
...
+          variation={ButtonVariation.ICON}
+          data-testid="close"
+          withoutBoxShadow
           onClick={() => {
-            trackEvent(FeatureActions.GetStartedPrevious, {
-              category: Category.FEATUREFLAG
-            })
-            onPrevious()
+            history.goBack()
```
knagurski, line 140: This may be a bit too simplistic. Could be worth opening this page in a new tab and seeing what behaviour it demonstrates. Also, do we store anything in the URL during the onboarding? I'm curious if it'll work in later stages of the onboarding flow. We may need to consider storing the URL of the place this was invoked from and going there and, if not recorded, go back to the Get Started page. Might be worth having a look at whether the version of React Router we use allows figuring out the previous page.

laura-mannus: Hi, although I agree it is simplistic it works completely so I don't see what is wrong with it? I have tested it for each stage of the onboarding flow, each empty state etc and it works as expected.

src/modules/75-cf/pages/onboarding/OnboardingDetailPage.tsx
laura-mannus: Cool, thank you - done

src/modules/75-cf/pages/onboarding/OnboardingDetailPage.tsx
laura-mannus: The reason for the `Icon` element is that it is custom. If we try to put it in the `Text` element as an icon it doesn't satisfy the IconName:

> Type '"cfLogo"' is not assignable to type 'IconName | undefined'.

laura-mannus: Haha good spot, resolved

src/modules/75-cf/pages/onboarding/OnboardingDetailPage.tsx
laura-mannus: If this works as expected, but a more in-depth way is needed, can we approve this PR and raise a separate ticket for this?

src/modules/75-cf/pages/onboarding/OnboardingDetailPage.module.scss
```diff
@@ -5,15 +5,32 @@
...
+  margin-left: var(--spacing-large) !important;
+  max-width: 250px !important;
+}
+
+.mainContent {
+  height: calc(var(--page-min-height) - var(--spacing-huge));
+  margin-left: var(--spacing-huge) !important;
 }
 
 .keep-on-top {
```
crystal-fong, line 29: this can be removed now

src/modules/75-cf/pages/onboarding/OnboardingDetailPage.tsx
```diff
@@ -111,116 +112,156 @@ export const OnboardingDetailPage: React.FC = () => {
...
-            setTestDone={setTestDone}
-            environmentIdentifier={selectedEnvironment.identifier}
-          />
-        )}
-      </Container>
-      <Divider />
-      <Layout.Horizontal spacing="small" padding="small">
+        <Layout.Horizontal
+          spacing="small"
+          style={{ justifyContent: 'flex-start', alignItems: 'center' }}
```
crystal-fong, line 124: since there's a css page that already exists, can you shift this into there?

knagurski: Actually, might as well use the `flex` prop

src/modules/75-cf/pages/onboarding/OnboardingDetailPage.tsx
```diff
@@ -111,116 +112,156 @@ export const OnboardingDetailPage: React.FC = () => {
...
+                flagInfo={selectedFlag as Feature}
+                language={language as PlatformEntry}
+                apiKey={apiKey as ApiKey}
+                testDone={testDone}
+                setTestDone={setTestDone}
+                environmentIdentifier={selectedEnvironment.identifier}
+              />
+            )}
+          </Container>
+          <Divider style={{ width: '100%' }} />
```
crystal-fong, line 221: here again - remove the inline styling

src/modules/75-cf/pages/onboarding/OnboardingDetailPage.tsx
Kira-Jennings: You might be able to utilise the history state which you can set when rerouting via `history.push()`
e.g. set something like `wasLogin: true` and then before doing `history.goBack()` check for this state and if true/present then tell it to go to a specific page, otherwise go back
image

src/modules/75-cf/pages/onboarding/OnboardingDetailPage.tsx
Kira-Jennings: +1 for flex prop

src/modules/75-cf/pages/onboarding/OnboardingDetailPage.tsx
knagurski: Might do the trick. When we navigate here when using the button, we store something and query it. We'll probably lose that when someone hits the refresh button, but it gets us 99% there! Good call @Kira-Jennings

laura-mannus: Thank you so much @Kira-Jennings this is a great call, will have a wee look now :)

src/modules/75-cf/pages/onboarding/OnboardingDetailPage.module.scss
laura-mannus: Nice spot thank you, meant to remove this!

src/modules/75-cf/pages/onboarding/OnboardingDetailPage.tsx
laura-mannus: Nice, thanks folks, why didn't I do this before?!

src/modules/75-cf/pages/onboarding/OnboardingDetailPage.tsx
laura-mannus: I have changed it here, but I didn't give it its own class in the css was because it is one attribute only used once.

src/modules/75-cf/components/GetStartedWithFF/GetStartedWithFF.tsx
```diff
@@ -34,6 +40,7 @@ export const GetStartedWithFF: React.FC<GetStartedWithFFProps> = ({ hidden }) =>
           size={ButtonSize.LARGE}
           text={getString('cf.featureFlags.getStartedWithFF')}
           onClick={() => {
+            setOnboardingBackLocation(true)
```
knagurski, line 43: Instead of storing a boolean, store the actual location to go back to. Since the preference store uses `localStorage`, that'll persist between sessions, meaning you can still get into a loop when they get logged out and try to access the onboarding flow by URL.

src/modules/75-cf/pages/onboarding/__test__/OnboardingDetailPage.test.tsx
```diff
@@ -30,6 +33,8 @@ const renderComponent = (): RenderResult => {
   )
 }
 
+jest.mock('framework/PreferenceStore/PreferenceStoreContext')
```
knagurski, line 36: Instead of mocking this and then mocking the implementation of the hook in each test, try just using a single mock or spy.

```suggestion
jest.mock('framework/PreferenceStore/PreferenceStoreContext', () => {
  usePreferenceStore: jest.fn().mockReturnValue({
    setPreference: jest.fn(),
    preference: true,
    clearPreference: jest.fn()
  })
})
```

or using a spy
```suggestion
import * as preferenceStoreContext from 'framework/PreferenceStore/PreferenceStoreContext'

// ...

jest.spyOn(preferenceStoreContext, 'usePreferenceStore').mockReturnValue({
  setPreference: jest.fn(),
  preference: true,
  clearPreference: jest.fn()
})

```

src/modules/75-cf/pages/onboarding/__test__/OnboardingDetailPage.test.tsx
```diff
@@ -43,8 +48,20 @@ describe('OnboardingDetailPage', () => {
   })
 
   test('OnboardingDetailPage empty state should be rendered properly', () => {
+    ;(usePreferenceStore as jest.Mock).mockImplementation(() => {
+      return {
+        setPreference: jest.fn(),
+        preference: true,
+        clearPreference: jest.fn
```
knagurski, line 55: Better not to mock in each test if it's always the same, but just a wee thing... when using `jest.fn`, you need to call it, not just reference it.
```suggestion
        clearPreference: jest.fn()
```

src/modules/75-cf/pages/onboarding/OnboardingDetailPage.tsx
```diff
@@ -111,116 +118,166 @@ export const OnboardingDetailPage: React.FC = () => {
...
-            if (selectedStep === STEP.SELECT_ENV_SDK) {
-              trackEvent(FeatureActions.SetUpYourApplicationVerify, {
-                category: Category.FEATUREFLAG
-              })
-            } else if (selectedStep === STEP.VALIDATE_FLAG) {
-              trackEvent(FeatureActions.TestYourFlagBack, {
-                category: Category.FEATUREFLAG
-              })
+            if (onboardingBackLocation) {
+              history.goBack()
```
knagurski, line 151: If we store the actual location we can then use `history.push` instead.

knagurski, line 151: Can we call `clearPreference` here to kill it off after we've used it?

src/modules/75-cf/pages/onboarding/__test__/OnboardingDetailPage.test.tsx
```diff
@@ -244,4 +269,33 @@ describe('OnboardingDetailPage', () => {
...
+      <TestWrapper
+        path="/account/:accountId/cf/orgs/:orgIdentifier/projects/:projectIdentifier/onboarding/detail"
+        pathParams={{ accountId: 'dummy', orgIdentifier: 'dummy', projectIdentifier: 'dummy' }}
+      >
+        <Router history={mockHistory}>
+          <OnboardingDetailPage />
+        </Router>
+      </TestWrapper>
+    )
+    userEvent.click(screen.getByTestId('close'))
```
knagurski, line 294: You should be able to ditch the memory history stuff if you switch to using an actual URL. In that case you can have a test for when you have a URL and when you don't and assert it goes to the URL when you do and goes to your safe place if you don't.

src/modules/75-cf/pages/onboarding/OnboardingDetailPage.module.scss
```diff
@@ -5,15 +5,32 @@
...
+}
+
+.mainContent {
+  height: calc(var(--page-min-height) - var(--spacing-huge));
+  margin-left: var(--spacing-huge) !important;
+}
+
+.buttonContainer {
+  display: flex;
+  align-items: flex-start;
```
Kira-Jennings, line 31: These can be put on the Layout element using this class instead

src/modules/75-cf/pages/onboarding/__test__/OnboardingDetailPage.test.tsx
```diff
@@ -55,6 +72,14 @@ describe('OnboardingDetailPage', () => {
   })
 
   test('Should be able to create a flag', async () => {
+    ;(usePreferenceStore as jest.Mock).mockImplementation(() => {
```
Kira-Jennings, line 75: is this colon a typo?

Kira-Jennings (dismissed):
Great work

src/modules/75-cf/pages/onboarding/OnboardingDetailPage.module.scss
laura-mannus: Thanks, have fixed

src/modules/75-cf/pages/onboarding/OnboardingDetailPage.tsx
laura-mannus: Thanks, have changed

laura-mannus: absolutely, thank you

src/modules/75-cf/pages/onboarding/__test__/OnboardingDetailPage.test.tsx
laura-mannus: Thanks, I have used the spy :) I was copying how we did it in other modules but this is far better

src/modules/75-cf/pages/onboarding/__test__/OnboardingDetailPage.test.tsx
laura-mannus: Thanks

src/modules/75-cf/pages/onboarding/__test__/OnboardingDetailPage.test.tsx
laura-mannus: This has been removed but it is standard :)

src/modules/75-cf/pages/onboarding/OnboardingDetailPage.tsx
```diff
@@ -47,12 +49,22 @@ export const OnboardingDetailPage: React.FC = () => {
   const [language, setLanguage] = useState<PlatformEntry>()
   const [apiKey, setApiKey] = useState<ApiKey>()
   const [selectedEnvironment, setSelectedEnvironment] = useState<EnvironmentResponseDTO | undefined>()
+  const [verified, setVerified] = useState(false)
   const [testDone, setTestDone] = useState(false)
+<<<<<<< HEAD
   const [verified, setVerified] = useState(false)
+=======
+
+>>>>>>> ef81b8e1a41a (feat: [FFM-5932]: Adding in new code from previous PR)
```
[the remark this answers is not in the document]
laura-mannus: This was on purpose as I was trying to fix something should be gone ow though :)

src/modules/75-cf/pages/onboarding/OnboardingDetailPage.tsx
```diff
@@ -114,119 +120,179 @@ export const OnboardingDetailPage: React.FC = () => {
 
   const { trackEvent } = useTelemetry()
 
+  const handleClose = useCallback(() => {
+    if (onboardingBackLocation) {
```
Kira-Jennings, line 124: So glad this worked :)

Kira-Jennings (approved):

src/modules/75-cf/pages/onboarding/OnboardingDetailPage.tsx
```diff
@@ -114,119 +120,179 @@ export const OnboardingDetailPage: React.FC = () => {
...
+          spacing="small"
+          flex={{ justifyContent: 'flex-start', alignItems: 'center' }}
+          padding={{ top: 'xxlarge' }}
+        >
+          <Text
+            font={{ variation: FontVariation.SMALL_BOLD }}
+            color={Color.GREY_800}
+            icon="cf-main"
+            iconProps={{ size: 30 }}
+            data-testid="cfLogo"
```
crystal-fong, line 163: the smallest nitpick - should it be `ffLogo`? Doesn't matter if you don't change it

crystal-fong (dismissed):
great work! 🦖

src/modules/75-cf/pages/onboarding/OnboardingDetailPage.tsx
laura-mannus: great point yeah, I called it that to match the name of the icon, I will address in my refactor ticket that will be coming because Ksenia is going to look at spacing etc. Thanks!

knagurski: Kill all `cf`! Lol. I fully support this course of action

src/modules/75-cf/components/GetStartedWithFF/GetStartedWithFF.tsx
```diff
@@ -34,6 +40,7 @@ export const GetStartedWithFF: React.FC<GetStartedWithFFProps> = ({ hidden }) =>
           size={ButtonSize.LARGE}
           text={getString('cf.featureFlags.getStartedWithFF')}
           onClick={() => {
+            setOnboardingBackLocation(history.location.pathname)
```
knagurski, line 43: So with this we lose some of the URL since it only gives you the path name, not the search string. `history.push` takes either a string or a full location object, so maybe the best idea is to JSON encode `history.location` and decode it when you're using it.

src/modules/75-cf/pages/onboarding/OnboardingDetailPage.tsx
```diff
@@ -47,18 +49,22 @@ export const OnboardingDetailPage: React.FC = () => {
...
+  const { preference: onboardingBackLocation, clearPreference: clearOnboardingBackLocation } =
+    usePreferenceStore<string>(PreferenceScope.USER, 'FF_ONBOARDING_LOCATION')
+
   const onNext = (): void => {
     if (currentStep !== totalSteps) {
       const nextStep = currentStep + 1
       setCurrentStep(nextStep)
     } else {
       history.push(routes.toCFFeatureFlags({ accountId: accountIdentifier, orgIdentifier, projectIdentifier }))
+      clearOnboardingBackLocation()
```
knagurski, line 67: Do this before navigating, just to make sure nothing weird happens.

src/modules/75-cf/pages/onboarding/OnboardingDetailPage.tsx
```diff
@@ -114,119 +120,179 @@ export const OnboardingDetailPage: React.FC = () => {
...
+    } else {
+      history.push({
+        pathname: routes.toCFConfigurePath({
+          orgIdentifier,
+          projectIdentifier,
+          accountId: accountIdentifier
+        })
+      })
+    }
+    clearOnboardingBackLocation()
```
knagurski, line 135: You can move this to before `history.push(onboardingBackLocation)`, just to make sure nothing weird happens. You don't need it for the `else` coz you know it's empty

src/modules/75-cf/pages/onboarding/__test__/OnboardingDetailPage.test.tsx
```diff
@@ -45,6 +54,11 @@ describe('OnboardingDetailPage', () => {
   test('OnboardingDetailPage empty state should be rendered properly', () => {
     renderComponent()
 
+    //Logo and [x] button
+    expect(screen.getByTestId('cfLogo')).toBeVisible()
+    expect(screen.getByText('common.purpose.cf.continuous')).toBeVisible()
```
knagurski, line 59: These two do the same thing now that the logo is part of the `Text`. Might as well ditch the `testId` one and remove it from the component.

src/modules/75-cf/pages/onboarding/__test__/OnboardingDetailPage.test.tsx
```diff
@@ -244,4 +258,27 @@ describe('OnboardingDetailPage', () => {
...
+  test('should navigate to the configure path url if you click the [x] icon and preference does not exist', async () => {
+    jest.spyOn(preferenceStoreContext, 'usePreferenceStore').mockReturnValue({
+      setPreference: jest.fn(),
+      preference: null,
+      clearPreference: jest.fn()
+    })
+    renderComponent()
+    userEvent.click(screen.getByTestId('close'))
+
+    expect(screen.findByTestId('location'))
```
knagurski, line 279: You can drop this as you're asserting it exists in the next line and this isn't an actual assertion.
```suggestion
```

src/modules/75-cf/pages/onboarding/__test__/OnboardingDetailPage.test.tsx
```diff
@@ -244,4 +258,27 @@ describe('OnboardingDetailPage', () => {
 
     expect(container).toMatchSnapshot()
   })
+
+  test('should navigate to the previous page if you click the [x] icon and preference exists', async () => {
+    renderComponent()
+    userEvent.click(screen.getByTestId('close'))
+
+    expect(screen.findByTestId('location'))
+    expect(screen.getByTestId('location')).toHaveTextContent(preferenceURL)
```
knagurski, line 267: You can ditch the first `expect` as it isn't actually doing anything

```suggestion
    expect(screen.getByTestId('location')).toHaveTextContent(preferenceURL)
```

src/modules/75-cf/pages/onboarding/OnboardingDetailPage.tsx
laura-mannus: Thanks, have moved

src/modules/75-cf/pages/onboarding/OnboardingDetailPage.tsx
laura-mannus: has been binned now :)

src/modules/75-cf/pages/onboarding/OnboardingDetailPage.tsx
laura-mannus: moved :)

src/modules/75-cf/pages/onboarding/__test__/OnboardingDetailPage.test.tsx
laura-mannus: have removed :)###

src/modules/75-cf/pages/onboarding/__test__/OnboardingDetailPage.test.tsx
laura-mannus: great point have removed

src/modules/75-cf/pages/onboarding/__test__/OnboardingDetailPage.test.tsx
laura-mannus: removed :)

Kira-Jennings (approved):

src/modules/75-cf/pages/onboarding/__test__/OnboardingDetailPage.test.tsx
```diff
@@ -30,13 +31,20 @@ const renderComponent = (): RenderResult => {
   )
 }
 
+const preferenceURL = '/testURL'
+
+jest.spyOn(preferenceStoreContext, 'usePreferenceStore').mockReturnValue({
+  setPreference: jest.fn(),
+  preference: JSON.stringify(preferenceURL),
+  clearPreference: jest.fn()
+})
```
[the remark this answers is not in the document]
laura-mannus: Have moved accordingly

crystal-fong (approved):

Kira-Jennings (approved):

PR #13361, opened 2023-01-09
feat: [CDS-44455]: Add node list pane for toggling visibility of matrix child nodes

rameshx:
### Summary

Adds a node list pane in the execution view for toggling the visibility of nodes when a looping strategy is used.
Shows a maximum of 8 nodes by default if no selection is made by the user. 
If any nodes are selected, only selected nodes are rendered in the graph.

#### Screenshots

https://user-images.githubusercontent.com/52284933/211321772-feebb219-df4f-4ff9-aab7-74a2ef3a3e5a.mov


src/modules/70-pipeline/components/ExecutionNodeList/CollapsedNodeStore.tsx
```diff
@@ -0,0 +1,61 @@
...
+export function CollapsedNodeProvider({ children }: { children: React.ReactNode }): JSX.Element {
+  const useReducerReturn = useReducer(reducer, null, createInitialState)
+
+  return <CollapsedNodeContext.Provider value={useReducerReturn}>{children}</CollapsedNodeContext.Provider>
+}
+
+export function useCollapsedNodeStore(): [CollapsedNodeState, React.Dispatch<CollapsedNodeAction>] {
+  const store = useContext(CollapsedNodeContext)
+  if (!store) {
+    throw new Error('useCollapsedNodeStore should be used within CollapsedNodeProvider')
```
PratyushGar, line 58: We can move this text to strings

PratyushGar (approved):

src/modules/70-pipeline/components/ExecutionNodeList/CollapsedNodeStore.tsx
rameshx: This is not displayed to the user, mostly useful during development if we accidentally try to use the hook without provider.

shaurya-harness (approved):

PR #14064, opened 2023-02-13
feat: [CDS-53471]: memoize getString

rameshx:
### Summary

Memoizes the `getString` function that is returned by the `useStrings` hook. With this change, `getString` can be safely included as a dependency when using `useEffect/useMemo/useCallback` etc.

Some hooks/components are incorrectly relying on the previous frequently changing `getString` function for rerendering/computing the next state. Among these components, the ones that had unit tests were easily identified because of failing tests and were fixed. The ones that don't have good coverage can only be identified manually once this is merged.


src/modules/75-cd/components/ServiceDetails/ServiceDetailsSummaryV2/ServiceDetailsEnvView.tsx
```diff
@@ -313,16 +332,24 @@ export function ServiceDetailsEnvView(props: ServiceDetailsEnvViewProps): React.
...
       ? renderCards.map((item, idx) => {
           const envInfo = item as EnvCardProps[]
           return (
             <Layout.Horizontal key={idx} className={css.envCardGrid}>
               {envSlide.map(i => {
+                if (!envInfo[i]) return null
                 return (
-                  envInfo[i] &&
-                  EnvCard(setSelectedEnv, setEnvId, setIsDetailsDialogOpen, setEnvFilter, envInfo[i], selectedEnv)
+                  <EnvCard
```
rameshx, line 344: `EnvCard` was previously used as a normal function causing a difference in the number of hooks rendered among rerenders. Updated to use this as a component.

udayvunnam (approved):

src/framework/strings/__tests__/Strings.test.tsx
```diff
@@ -150,6 +150,19 @@ describe('useString tests', () => {
...
+      const wrapper = ({ children }: React.PropsWithChildren<unknown>): React.ReactElement => (
+        <StringsContext.Provider value={value as any}>{children}</StringsContext.Provider>
+      )
+      const { result, rerender } = renderHook(useStrings, { wrapper })
+      const prevGetString = result.current
+
+      rerender()
+      const nextGetString = result.current
+
+      expect(prevGetString).toEqual(nextGetString)
```
vkbansal-harness, line 164: 🔥

src/modules/75-cd/components/ServiceDetails/ServiceDetailsSummaryV2/ServiceDetailsEnvView.tsx
vkbansal-harness: We should add an ESLint rule for banning such uses

vkbansal-harness (approved):

sunnykesh-harness (approved):

PR #14173, opened 2023-02-21
fix: [CDS-52327]: Template inputs api moved to template details compo…

samarth-sharma31:
…nent to avoid new call when tab is switched

### Summary

Template inputs api call has been moved to template details component when the inputs are displayed in the template right drawer. This reduces the number of template api calls to fetch the inputs, especially when the tab is switched between inputs and yaml screen

#### Screenshots

NA

https://user-images.githubusercontent.com/96409598/220318532-77922303-7ffa-48e6-a71b-cc95ca8d2e56.mov

#### PR Checklist

<details>

</details>

<details>

- Jest: `retrigger jest`
- ESLint: `retrigger eslint`
- Sonar: `retrigger sonar`
- Build: `retrigger build`
- Rebase: `trigger rebase`
</details>


src/modules/72-templates-library/components/TemplateInputs/TemplateInputs.tsx
```diff
@@ -94,11 +111,13 @@ export const TemplateInputs: React.FC<TemplateInputsProps> = ({ template, storeM
   const branch =
     (template as TemplateSummaryResponse).gitDetails?.branch || (template as NGTemplateInfoConfigWithGitDetails).branch
 
+  const isTemplateInputSetAlreadyFetched = !isEmpty(templateInputSetFetchParams)
```
AshwinBhatkal-Harness, line 114: When would this be empty? I'm thinking, let's render the templateInputs component only after the data has loaded. This way we can totally avoid the useGetTemplateInputSetYaml in this component

samarth-sharma31: Actually template inputs component is used standalone also - when we show template inputs inside template studio (check image for reference) - so that's why this component always had logic for fetching template inputs -> useGetTemplateInputSetYaml - So this will be empty when this data has not been fetched by the parent component - which is the above mentioned scenario.

Screenshot 2023-02-22 at 1 11 49 AM

So, the approach that i have taken here - if template input set yaml is fetched by parent component, then the api here in child component is marked as lazy and if it has not been fetched and passed by the parent component - then child component works the way it used to work before - this solve the problem at both places.
let me know if we need to connect for a call to discuss this further. Thanks!

AshwinBhatkal-Harness: Got it. So in this case, we could keep break it into 2 components
1. Pure template inputs - gets the data as props and renders it
2. Wrapper around template inputs - API call is made and passes data to component 1

Component 1 can be used across the factory and inside the wrapper
Component 2 will be used for standalone use cases

So we could have TemplateInputs and TemplateInputsStandAlone
TemplateInputs would render data
TemplateInputsStandAlone would have TemplateInputs as child

The concern that I am trying to address is -> "I have already made a template input set yaml call, the internal component may also make the same call/ at least has code for such a use case. Since the current TemplateInputs component is used across several files, it might be confusing as well"

AshwinBhatkal-Harness: If not following the above, we can have some comments written around the below piece of code. This will make us future error proof

AshwinBhatkal-Harness (approved):

Sagar-Rajput7 (approved):

PR #14178, opened 2023-02-21
fix: [FFM-6921]: Use flag identifier in onboarding examples

knagurski:
This PR replaces the usage of flag name in onboarding examples with flag identifier to fix a bug whereby entering a dash or using capital letters in the name would result in the name and identifier not being equal and the example not being able to successfully evaluate the flag.


src/modules/75-cf/strings/strings.en.yaml
```diff
@@ -1569,7 +1569,7 @@ onboarding:
         apiKey = '{{apiKey}}''
 
         # Flag Name
```
crystal-fong, line 1571: does this need to be `Flag Identifier`?

knagurski: Good eye! Fixed now

knagurski:
Force merged due to unrelated flaky test failure

PR #14354, opened 2023-03-01
feat: [FFM-6890]: Improving audit wording for add rule

laura-mannus:
### Summary

**AS A** FF admin user
**I WANT** to see what rule was added in plain English
**SO THAT** I can understand how this impacts targeting without needing to understand the YAML diff

I have updated the add rule wording for both target and target groups to read more clearly. There was a limitation for the remove rule on target groups as we don't have the data for that. It will remain "Rule removed" however the log for Target is more verbose.

#### Screenshots

Before

image

After

image

#### PR Checklist

<details>

</details>

<details>

- Jest: `retrigger jest`
- ESLint: `retrigger eslint`
- Sonar: `retrigger sonar`
- Build: `retrigger build`
- Rebase: `trigger rebase`
</details>


src/modules/75-cf/strings/strings.en.yaml
```diff
@@ -391,10 +391,11 @@ auditLogs:
       bucketBy: Default serve updated for bucket by '{{bucketBy}}'
       variation: Default serve updated with variation '{{variation}}'
     addTargetsToVariationTargetMap: Target '{{target}}' was added for Variation '{{variation}}'
+    addTargetToFlagsVariationTargetMap: Target '{{target}}' was added for Variation '{{variation}}'
```
laura-mannus, line 394: is there a better way to achieve this than having two with the same value? I tried using `addTargetsToVariationTargetMap ` for both cases but this felt incorrect.

crystal-fong (dismissed):
Should this be reviewed first, then if it's all good apply the changes into https://github.com/wings-software/ff-ui/pull/60/files ?

src/modules/75-cf/strings/strings.en.yaml
knagurski: I *think* this could do the same job without duplication.
```suggestion
    addTargetToFlagsVariationTargetMap: {{ $.cf.auditLogs.events.addTargetsToVariationTargetMap }}
```

Not very pretty though. You could just use the same string id for both, but change it to be a little more generic... maybe `targetAddedForVariation`

src/modules/75-cf/components/AuditLogs/AuditLogsUtils.ts
```diff
@@ -47,9 +46,11 @@ export function translateEvents(
 
       case 'addRule':
         message = getString('cf.auditLogs.events.addRule', {
-          clauses: Parameters?.clauses?.map((clause: Record<string, string>) => clause?.attribute || '').join(', ')
+          targetGroup: Parameters?.clauses?.map((clause: Record<string, string>) => clause?.values || '').join(', '),
+          variation: Parameters.variation
```
knagurski, line 50: I'm not sure this will work for a Percentage Rollout. In their case `Parameters.variation` won't exist and `Parameters.serve` will be in its place. 
Screenshot 2023-03-02 at 11 27 31

laura-mannus: I added some logic in there, what do you think of?
image

Should the wording need updated as well?

src/modules/75-cf/strings/strings.en.yaml
laura-mannus: Have used the latter suggestion

knagurski: Yeah, I think that's the better idea. Referencing is good in some situations, but it feels kinda forced here.

src/modules/75-cf/components/AuditLogs/AuditLogsUtils.ts
laura-mannus: Looks like this now 
image

src/modules/75-cf/components/AuditLogs/AuditLogsUtils.ts
```diff
@@ -46,10 +45,19 @@ export function translateEvents(
...
+        Parameters.variation
+          ? (message = getString('cf.auditLogs.events.addRule', {
+              targetGroup: Parameters?.clauses
+                ?.map((clause: Record<string, string>) => clause?.values || '')
+                .join(', '),
+              variation: Parameters.variation
+            }))
+          : (message = getString('cf.auditLogs.events.addPercentageRollout', {
+              targetGroup: Parameters?.clauses?.map((clause: Record<string, string>) => clause?.values || '').join(', ')
+            }))
```
knagurski, line 57: Generally avoid assignments in ternaries... they're not super obvious and are usually better written as standard `if`/`else` statements
```suggestion
        if (Parameters.variation) {
          message = getString('cf.auditLogs.events.addRule', {
            targetGroup: Parameters?.clauses
              ?.map((clause: Record<string, string>) => clause?.values || '')
              .join(', '),
            variation: Parameters.variation
          })
        } else {
          message = getString('cf.auditLogs.events.addPercentageRollout', {
            targetGroup: Parameters?.clauses?.map((clause: Record<string, string>) => clause?.values || '').join(', ')
          })
        }
```

Or you could assign the output of the ternary
```suggestion
        message = Parameters.variation
          ? getString('cf.auditLogs.events.addRule', {
              targetGroup: Parameters?.clauses
                ?.map((clause: Record<string, string>) => clause?.values || '')
                .join(', '),
              variation: Parameters.variation
            })
          : getString('cf.auditLogs.events.addPercentageRollout', {
              targetGroup: Parameters?.clauses?.map((clause: Record<string, string>) => clause?.values || '').join(', ')
            })
```

knagurski:
Great work, just that ternary thing and I think we're good to go!

src/modules/75-cf/components/AuditLogs/AuditLogsUtils.ts
laura-mannus: Ok, I prefer the tidiness of the ternary so I have moved the assignment to the output 👍

knagurski (approved):
🔥 great work

crystal-fong (approved):
nice work 🦖

PR #14591, opened 2023-03-10
feat: [FFM-7180]: ChildComponentMounter

knagurski:
This PR creates a new `ChildComponentMounter` component and extracts the existing `ChildAppMounter` to use this new component. The new component does not pass any hooks, components or utilities to the child app by default. The modified `ChildAppMounter` component is vastly simplified to use the new component under-the-hood instead of providing the `Suspense`, loading indicator or error states.

The new `ChildComponentMounter` is intended to be used in cases where only a portion of an app needs to be mounted and where the standard components, hooks and utilities are overkill.


src/microfrontends/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@harness/microfrontends",
-  "version": "1.42.0",
+  "version": "1.43.0",
```
abhinavrastogi-harness, line 3: i believe this change be logically a major change... just so everyone is clear they need to test their MFEs integrations when they start using this.
let's go to `2.0.0`

abhinavrastogi-harness (approved):

bhavya-sinha (approved):

src/microfrontends/package.json
knagurski: I'm torn. Really this is entirely backwards compatible and outwardly just introduces a new component (for use within NGUI) and some new additional types (for use within MFEs). There is no breaking change, so I'd be inclined to follow semver on the numbering. MFE owners should ensure their apps still work with this new version, but unless there's a bug, there's no action needed unless they want to use the new mounter

PR #14628, opened 2023-03-13
feat: [FFM-6596]: Track pagination and filtration in URL

knagurski:
This PR switches page number, search term and filtration from `setState` to `setQueryParamsState`.


src/modules/75-cf/pages/feature-flags/FeatureFlagsPage.tsx
```diff
@@ -5,7 +5,7 @@
  * https://polyformproject.org/wp-content/uploads/2020/06/PolyForm-Shield-1.0.0.txt.
  */
 
-import React, { useEffect, useMemo, useRef, useState, useCallback, ReactElement } from 'react'
+import React, { ReactElement, useCallback, useEffect, useMemo, useRef, useState } from 'react'
```
knagurski, line 8: Most of this stuff is just rearranging imports. My IDE did it, but it's kinda nice IMO 😄

src/modules/75-cf/pages/feature-flags/FeatureFlagsPage.tsx
```diff
@@ -418,9 +419,11 @@ const FeatureFlagsPage: React.FC = () => {
   const history = useHistory()
   const { activeEnvironment: environmentIdentifier, withActiveEnvironment } = useActiveEnvironment()
   const searchRef = React.useRef<ExpandingSearchInputHandle>({} as ExpandingSearchInputHandle)
```
crystal-fong, line 421: NBD but `useRef` has been destructured so could just use that

src/modules/75-cf/pages/feature-flags/FeatureFlagsPage.tsx
```diff
@@ -614,8 +617,8 @@ const FeatureFlagsPage: React.FC = () => {
 
   const onSearchInputChanged = useCallback(
     name => {
-      setSearchTerm(name)
-      setPageNumber(0)
+      defer(() => setSearchTerm(name))
+      defer(() => setPageNumber(0))
```
crystal-fong, line 621: if I remember correctly `useQueryParamsState` acts like a useState but with some bells and whistles. So this `defer` to make sure onSearchInputChanged gets called first before `useQueryParamsState` ? JW

src/modules/75-cf/pages/feature-flags/FeatureFlagsPage.tsx
knagurski: There is _SOOO_ much I want to change in this file!

src/modules/75-cf/pages/feature-flags/FeatureFlagsPage.tsx
knagurski: So, there's an issue with `useQueryParamsState` in that if you set multiple bits of state synchronously using this, it loses some of them and only a wee bit actually takes effect. By deferring for a ms, it makes it async and allows multiple set state calls to actually work. I would've fixed this at the source, but it broke a bunch of tests across the app, so figured I'd keep it local for now. `defer` is roughly equiv to `setTimeout(thing, 1)`.

src/modules/75-cf/pages/feature-flags/FeatureFlagsPage.tsx
crystal-fong: "if I had the time I would write it from scratch!"

knagurski: command-a, backspace, enter sensible stuff, save, relax

crystal-fong (approved):

knagurski:
Had to force merge as unrelated Cypress tests kept failing due to flakiness.

PR #14660, opened 2023-03-14
feat: [FFM-6595]: Track targets groups pagination and filtration in URL

knagurski:
This PR tracks pagination and filtration of the Target Groups listing page in the URL.


src/modules/75-cf/pages/target-management/segments/SegmentsPage.tsx
```diff
@@ -72,11 +74,10 @@ export const SegmentsPage: React.FC = () => {
   const history = useHistory()
   const onSearchInputChanged = useCallback(
     name => {
-      setSearchTerm(name)
-      setPageNumber(0)
-      refetchSegments({ queryParams: { ...queryParams, name, pageNumber: 0 } as GetAllSegmentsQueryParams })
```
crystal-fong, line 77: where does the refetching happens now it's removed here?

knagurski: The refetch happens automatically coz the search term's state change triggers the query to fire again. If it wasn't in the query params of the original call, we'd have to call the refetch manually. As it was it calling the refetch unnecessarily

crystal-fong: okie cool

crystal-fong (approved):

laura-mannus (approved):

PR #14787, opened 2023-03-21
fix: [FFM-7234]: Display missing permission text

knagurski:
This PR adds labels for FF-specific permissions against environments. This should be considered a short/medium term fix for a display issue while FF reconsiders its RBAC setup.


tan-nhu (approved):

devinder-harness (approved):

kapilkumawat86 (approved):

src/modules/70-pipeline/RouteDestinations.tsx
```diff
@@ -124,7 +124,12 @@ RbacFactory.registerResourceTypeHandler(ResourceType.ENVIRONMENT, {
     [PermissionIdentifier.VIEW_ENVIRONMENT]: <String stringID="rbac.permissionLabels.view" />,
     [PermissionIdentifier.EDIT_ENVIRONMENT]: <String stringID="rbac.permissionLabels.createEdit" />,
     [PermissionIdentifier.DELETE_ENVIRONMENT]: <String stringID="rbac.permissionLabels.delete" />,
-    [PermissionIdentifier.RUNTIMEACCESS_ENVIRONMENT]: <String stringID="rbac.permissionLabels.access" />
+    [PermissionIdentifier.RUNTIMEACCESS_ENVIRONMENT]: <String stringID="rbac.permissionLabels.access" />,
+
+    // these are required for the FF module to restrict flag toggling and editing based on permissions
+    // against specific environments
+    [PermissionIdentifier.TOGGLE_FF_FEATUREFLAG]: <String stringID="rbac.permissionLabels.toggleFlag" />,
```
abhinavrastogi-harness, line 131: Ideally module-specific logic should not be added to `70-pipeline`. but I don't see a way around this right now because `environments` are shared 🤔

abhinavrastogi-harness (approved):

sunnykesh-harness (approved):

src/modules/70-pipeline/RouteDestinations.tsx
knagurski: I know. It feels wrong to have to do this in Pipelines, but I couldn't see any other way either. Very open to ripping this out if we do find a better way.

PR #14947, opened 2023-03-29
feat: [CDS-58013]: Add link to IP whitelisting page

ishantmehtaharness:
Gitops PLG does not have any jest for now. So we can FM this


ishantmehtaharness:
retrigger titlecheck

src/modules/75-cd/strings/strings.en.yaml
```diff
@@ -631,7 +631,9 @@ getStartedWithCD:
   clusterSpecInfo: This specification is sufficient for you to host a basic hello world application
   clusterspec: Cluster Specifications
   managedCluster: Make use of our managed cluster to facilitate your deployment if you’re trying out Harness CD-GitOps for the first time, this will expire after X days! Learn more
-  ipWhitelist: You will be required to Set up IP Whitelisting for Harness to connect to your cluster
+  ipWhitelist1: 'You will be required to '
+  ipWhitelist2: ' for Harness to connect to your cluster'
```
Sagar-Rajput7, line 635: NIT: just curious here, if we support <a> tag as the 2nd prop in getString, if yes then we can replace these 2 with one having  a `{link}`.

Sagar-Rajput7 (approved):

tan-nhu (approved):

PR #15230, opened 2023-04-18
feat: [CDS-58620]: PLG - Do. not enable "Next Configure" until agent …

ishantmehtaharness:
coverage check is failing - since we dont have jest testing for PLG flow currently


sahithikolichala (dismissed):

rameshx (approved):

src/modules/75-cd/pages/get-started-with-cd/GitOpsAgent/AgentProvision.tsx
```diff
@@ -80,6 +82,14 @@ export const AgentProvision = ({
     }
   }, [agentCreateError])
 
+  let calledOnce = false
```
tan-nhu, line 85: This does not seem correct as the component can be run again and again, every time this var is false. You'll need to use useState or useRef instead

ishantmehtaharness: Hey @tan-nhu , you are correct.
I tried debugging this.
And seems like this effect will only run when these dependencies change `[isHealthy, data?.health?.lastHeartbeat]`, which is working fine.

Also, your suggestion is on the good practices, I have made the changes you suggested. Thank You :)

ishantmehtaharness: I have made the changes

rameshx (approved):

sahithikolichala (approved):

tan-nhu (approved):

PR #15363, opened 2023-04-24
fix: [FFM-7609]: Await async and avoid pre-mature exception handling

knagurski:
This PR fixes an issue where the success message was immediately shown before the request completed and any errors would be displayed, but not halt the process.


Kira-Jennings (approved):

laura-mannus (approved):
nice one

PR #15405, opened 2023-04-25
fix: [CDS-59192]: enhance duplicate identifier detection

rameshx:
### Summary

Updated the duplicate identifier check that runs when the studio view switches from YAML to the visual view. 
The new function generates FQN-like paths to check for duplicates. This change is mainly done because the pipeline YAML can have repeated identifiers as long as their paths are unique.

https://user-images.githubusercontent.com/52284933/234381025-c6945d0e-7d14-42b0-9977-5eb60036555e.mov


rameshx:
@YounixM This needs a force merge as branch coverage for `PipelineActions.ts` is too low

PratyushGar (approved):

YounixIO (approved):

PR #15622, opened 2023-05-08
fix: [PL-38495]: fix error when searchTerm starts with a number

rameshx:
### Summary

Fixes page crash caused when searchTerm starts with a number.

#### Screenshots

https://user-images.githubusercontent.com/52284933/236809693-867bab3b-9dd1-418b-ab02-ff2212da4123.mov


src/modules/10-common/hooks/useQueryParams.ts
```diff
@@ -85,10 +85,19 @@ export const useQueryParamsOptions = <Q, DKey extends keyof Q>(
...
     () => ({
       decoder: queryParamDecodeAll(),
-      processQueryParams: (params: Q) =>
-        assignWith({ ...params }, defaultParamsRef.current, (objValue, srcValue) =>
+      processQueryParams: (params: Q) => {
+        const processedParams = { ...params }
+
+        // if searchTerm is '123', queryParamDecodeAll converts it to the number 123
+        // but searchTerm should remain a string
+        if (!isNil(get(processedParams, 'searchTerm'))) {
```
mayankVermaHarness, line 93: can we use something like:

`processedParams?.searchTerm?.toString()`

rameshx: yes we can use optional chaining, but `get` does the same, this hook is mostly using lodash so it is consistent

mayankVermaHarness: we can get rid of the `If` condition with `optional chaining` but okay, approving

mayankVermaHarness (approved):

pankajsingh-harness (approved):

PR #15947, opened 2023-05-24
fix: [FFM-8002]: Always include FF routes

knagurski:
This PR ensures that the FF-related routes are always loaded, regardless of license.


crystal-fong (approved):

knagurski:
Force merged to get to hotfix asap.

PR #15958, opened 2023-05-24
fix: [CDS-56532]: Maven, Nuget and Npm support added for github packages

samarth-sharma31:
#### Screenshots

NA

#### PR Checklist

<details>

</details>

<details>

- Jest: `retrigger jest`
- ESLint: `retrigger eslint`
- Sonar: `retrigger sonar`
- Build: `retrigger build`
- Rebase: `trigger rebase`
</details>


src/modules/70-pipeline/components/ArtifactsSelection/ArtifactInterface.ts
```diff
@@ -65,6 +65,12 @@ export enum TagTypes {
   Value = 'value',
   Regex = 'regex'
 }
+
+export enum PackageSourceTypes {
```
pankajsingh-harness, line 69: No need to define this enum as `packageSource?: 'org' | 'user'`. Just use string to pass the value. It will not accept any other values

src/modules/70-pipeline/components/ArtifactsSelection/ArtifactUtils.ts
```diff
@@ -526,6 +527,12 @@ const getVersionValues = (
   AmazonMachineImageInitialValuesType => {
   const formikInitialValues = {
     versionType: specValues?.version ? TagTypes.Value : TagTypes.Regex,
+    packageSource:
+      specValues?.packageType === 'maven'
+        ? specValues?.user
+          ? PackageSourceTypes.User
```
pankajsingh-harness, line 533:
```suggestion
          ? 'user'

File: src/modules/70-pipeline/components/ArtifactsSelection/ArtifactUtils.ts [hunk 3]
```diff
@@ -526,6 +527,12 @@ const getVersionValues = (
...
+          : PackageSourceTypes.Org
```
pankajsingh-harness, line 534:
```suggestion
          : 'org'
```

src/modules/70-pipeline/components/ArtifactsSelection/ArtifactInterface.ts
samarth-sharma31: I feel it is better to use enums for values which are defined properly - Reasons - >

1. Makes code more maintainable - in the sense that is any changes need to be done to the string - will not have to search it explicitly  - usage of this Enum makes it very easier to work with
2. Reduces chances of using typo

Lets say in future  if some one needs to add more checks when package source is user -it is very easier to check occurances of PackageSourceTypes rather than searching for users.

pankajsingh-harness: @samarth-sharma31 I agree with this. But here we are making 2 sources of truth. If we have the proper Discriminating Unions, TypeSccript will give proper suggestions and errors and also does not generate any code. Enum produces js code

samarth-sharma31: Yes @pankajsingh-harness , agreed, there is a trade off here. 
Okay so in order to combine the benefits of both the approaches that we have discussed - i think i should use objects as const --> This approach will give us all the benefits we want, without generating extra code.

Please check this article i just found --> https://dev.to/ivanzm123/dont-use-enums-in-typescript-they-are-very-dangerous-57bh#:~:text=One%20of%20the%20reasons%20why,and%20performance%20of%20the%20app.

This also discusses the same problem that you mentioned.

samarth-sharma31: As discussed, will be updating it in a follow up PR

sonaligoyal-harness (approved):

chetanvekariya-harness (approved):

PR #16139, opened 2023-06-02
feat: [CDS-70491]: add support for search and pagination to input set selector

rameshx:
This change updates the input set selector to support API-based search and pagination

The video below uses `5` as the page size as an example. The actual page size used is 100 to be consistent with the old size.

https://github.com/harness/harness-core-ui/assets/52284933/81ee5ac5-5617-4240-a5c1-9f34692a5010


ishantmehtaharness (approved):
Thanks @rameshx for fixing it quickly. I see this was raised last week by Suma.

pankajsingh-harness (approved):

udayvunnam (approved):

PR #16420, opened 2023-06-19
feat: [FFM-7936]: listing archived flags

crystal-fong:
Users are able to archive a flag
The 'Archived' card will appear
Users can click the Archived card and view the list of archived flag
Users can restore an archived flag
The restored flag will appear back in the normal flag listing

When users not delete a flag, it will be permanently deleted

#### Screenshots

https://github.com/harness/harness-core-ui/assets/95680496/7e6158e1-cc9b-419c-9120-c2369179105a

#### PR Checklist

<details>

</details>

<details>

- Jest: `retrigger jest`
- ESLint: `retrigger eslint`
- Sonar: `retrigger sonar`
- Build: `retrigger build`
- Rebase: `trigger rebase`
</details>


src/services/cf/index.tsx
```diff
@@ -3655,11 +3657,16 @@ export interface DeleteFeatureFlagQueryParams {
    * Git commit message
    */
   commitMsg?: string
+  /**
+   * Permanently deletes the the feature flag
+   */
+  forceDelete?: boolean
```
knagurski, line 3663: Does this mean that the backend now archives by default? That's a breaking change. Is it in production yet?

crystal-fong: the BE logic should also be behind a FF

src/modules/75-cf/components/FlagArchiving/__tests__/useRestoreFlagDialog.test.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+    restoreFlagMutate.mockRejectedValue({ message: error })
+
+    renderComponent({ refetchFlags: refetchFlagMock })
+
+    await userEvent.click(screen.getByRole('button', { name: openRestoreDialogBtn }))
+    userEvent.click(screen.getByRole('button', { name: 'cf.featureFlags.archiving.restore' }))
+
+    await waitFor(() => {
+      expect(screen.getAllByText(error)[0]).toBeInTheDocument()
```
crystal-fong, line 90: some weird mocking causes the error toaster to appear more than once but I'm too sick of this PR :D

src/modules/75-cf/components/FlagOptionsMenuButton/FlagOptionsMenuButton.tsx
```diff
@@ -92,36 +99,71 @@ const FlagOptionsMenuButton: FC<FlagOptionsMenuButtonProps> = ({
...
+            permission: PermissionIdentifier.DELETE_FF_FEATUREFLAG
+          },
+          ...planEnforcementProps
+        }
+      ]
 
-    if (noEdit) {
-      opts.shift()
+      if (noEdit) {
+        optionItems.shift()
```
d1wilko, line 152: I appreciate you didn't write the initial code here - but using the shift function assumes edit will always be the first element... if someone changed the code above and added an option in first this would break

perhaps this should be changed to not include edit option in the first place if noEdit is true?

src/modules/75-cf/pages/feature-flags/components/FlagTableFilters.tsx
```diff
@@ -67,17 +67,31 @@ export const featureFlagFilters = (features: Features | null): Array<FilterProps
...
   let filters = featureFlagFilters(features)
 
   // remove 'Active Flags' filter card if feature flag FFM_3938_STALE_FLAGS_ACTIVE_CARD_HIDE_SHOW is not enabled
   if (!DISPLAY_ACTIVE_FILTER) {
     filters = filters.filter(filter => filter.label !== 'cf.flagFilters.active')
   }
 
+  // do not show the Archived card if there are no flags that are archived and if Archiving FF is toggled OFF
+  if (!ARCHIVING_FEATURE_FLAGS || !features?.featureCounts?.totalArchived) {
+    filters.pop()
```
d1wilko, line 93: again this assumes ARCHIVED will always be the last element... it could break if new options are added

src/modules/75-cf/pages/feature-flags/components/__tests__/FlagTableFilters.test.tsx
```diff
@@ -40,7 +44,7 @@ const permanentFlagsFilter = {
 
 describe('FlagTableFilters', () => {
   beforeEach(() => {
-    flagEnabled = true
+    ;(flagEnabled = true), (archivingFFEnabled = false)
```
d1wilko, line 47: is the semicolon at the start here right?

d1wilko (changes requested):
I don't understand feature flags lol - some small suggestions, but coming together nicely

src/modules/75-cf/pages/feature-flags/components/__tests__/FlagTableFilters.test.tsx
crystal-fong: yeaaah prettier added this automatically

src/modules/75-cf/components/FlagDetailsOptionsMenuButton/FlagDetailsOptionsMenuButton.tsx
```diff
@@ -82,7 +82,7 @@ const FlagDetailsOptionsMenuButton: FC<FlagDetailsOptionsMenuButtonProps> = ({
   const { openDialog } = useArchiveFlagDialog({
     flagData: featureFlag,
     queryParams,
-    deleteFeatureFlag,
+    archiveFlag: deleteFeatureFlag,
```
laura-mannus, line 85: any way we could rename to just be `archiveFlag` since we aren't using the delete term anymore

src/modules/75-cf/components/FlagOptionsMenuButton/FlagOptionsMenuButton.tsx
```diff
@@ -70,11 +71,17 @@ const FlagOptionsMenuButton: FC<FlagOptionsMenuButtonProps> = ({
     onSuccess: () => refetchFlags?.()
   })
 
-  const { openDialog } = useArchiveFlagDialog({
+  const { openDialog: openArchiveDialog } = useArchiveFlagDialog({
     flagData,
     queryParams,
     refetchFlags,
-    deleteFeatureFlag: deleteFlag
+    archiveFlag: deleteFlag
```
laura-mannus, line 78: same here, just seems odd we refer to delete here. Can we just have archive?

crystal-fong: the delete functionality still exists, the archive and delete functionality is just differentiated by a query parameter so if I change it here, hard delete would look like I'm archiving

src/modules/75-cf/components/FlagArchiving/useRestoreFlagDialog.tsx
```diff
@@ -0,0 +1,62 @@
...
+  const { handleResponseError } = useResponseError()
+
+  const queryParams = useMemo(
+    () => ({
+      identifier: flagData.identifier,
+      environmentIdentifier: flagData.envProperties?.environment,
+      ...restoreQueryParams
+    }),
+    [flagData.envProperties?.environment, flagData.identifier, restoreQueryParams]
+  )
```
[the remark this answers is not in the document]
crystal-fong: I've removed the useMemo now and queryParams is no longer a function

src/modules/75-cf/components/FlagOptionsMenuButton/FlagOptionsMenuButton.tsx
crystal-fong: that's a fair point, I've updated it now ty

src/modules/75-cf/components/FlagArchiving/__tests__/useArchiveFlagDialog.test.tsx
```diff
@@ -6,19 +6,20 @@ import mockFeature from '@cf/components/EditFlagTabs/__tests__/mockFeature'
 import type { Feature } from 'services/cf'
 import useArchiveFlagDialog, { ArchiveDialogProps } from '../useArchiveFlagDialog'
 
-const deleteFeatureFlagMock = jest.fn()
+const archiveFeatureFlagMock = jest.fn()
```
knagurski, line 9: Looks like these mocks aren't cleared between tests, so running the tests in different order could give different results. Can we put in a `jest.clearAllMocks` in a `beforeEach` block?

src/modules/75-cf/pages/feature-flags/components/__tests__/FlagTableFilters.test.tsx
```diff
@@ -143,4 +147,62 @@ describe('FlagTableFilters', () => {
...
+    expect(getByTestId(filterCards[0], 'filter-label')).toHaveTextContent('cf.flagFilters.allFlags')
+    expect(getByTestId(filterCards[1], 'filter-label')).toHaveTextContent('cf.flagFilters.enabled')
+    expect(getByTestId(filterCards[2], 'filter-label')).toHaveTextContent('cf.flagFilters.permanent')
+    expect(getByTestId(filterCards[3], 'filter-label')).toHaveTextContent('cf.flagFilters.recentlyAccessed')
+    expect(getByTestId(filterCards[4], 'filter-label')).toHaveTextContent('cf.flagFilters.potentiallyStale')
+
+    expect(screen.queryByText('cf.flagFilters.archivedFlags')).not.toBeInTheDocument()
+  })
+
+  test('It should not show Archived card if there are no archived flags', async () => {
```
knagurski, line 185: Out of curiosity, what happens if you are in the archived flags view with a single archived flag and then you restore it? If archived flags disappears, then we'd need to make sure the user is redirected. Would it maybe make more sense to always show the Archived Flags card like any other card?

src/modules/75-cf/pages/feature-flags/components/__tests__/FlagTableFilters.test.tsx
```diff
@@ -143,4 +147,62 @@ describe('FlagTableFilters', () => {
...
+    expect(getByTestId(filterCards[2], 'filter-label')).toHaveTextContent('cf.flagFilters.permanent')
+    expect(getByTestId(filterCards[3], 'filter-label')).toHaveTextContent('cf.flagFilters.recentlyAccessed')
+    expect(getByTestId(filterCards[4], 'filter-label')).toHaveTextContent('cf.flagFilters.potentiallyStale')
+    expect(getByTestId(filterCards[5], 'filter-label')).toHaveTextContent('cf.flagFilters.archivedFlags')
+  })
+
+  test('It should not show Archived card if FFM_7921_ARCHIVING_FEATURE_FLAGS is toggled OFF', async () => {
+    flagEnabled = false
+    archivingFFEnabled = false
+    renderComponent()
```
knagurski, line 171: It's usually a better idea to pass these into the `renderComponent` function instead of using global variables, otherwise you can have unexpected results and tests will fail if you reorder them.

src/modules/75-cf/pages/feature-flags/components/__tests__/FlagTableFilters.test.tsx
knagurski: That's the downside of using JS without semicolons... sometimes they're needed and can't reliably be auto-injected at runtime. Though in this case a line break between seems fine in my IDE.
```suggestion
    flagEnabled = true
    archivingFFEnabled = false
```

src/modules/75-cf/pages/feature-flags/components/__tests__/FlagTableFilters.test.tsx
crystal-fong: image
it'll look like ^

knagurski: So it just kinda orphans it 🤔

src/modules/75-cf/components/FlagArchiving/__tests__/useArchiveFlagDialog.test.tsx
knagurski: Ideally it should just be `jest.fn()` on line 32 and then you create and pass in only in the tests you actually need it for (i.e. only those you assert on it)

src/modules/75-cf/components/FlagArchiving/__tests__/useRestoreFlagDialog.test.tsx
knagurski: It's not that it appears more than once, it's that it doesn't disappear between tests. We have a wee bit of code in FF-UI that kills it off between tests.

src/modules/75-cf/pages/feature-flags/components/__tests__/FlagTableFilters.test.tsx
knagurski: We don't remove any other filters based on availability of data. Could be worth confirming if that's really what we want to do here.

src/modules/75-cf/pages/feature-flags/components/FlagTableFilters.tsx
```diff
@@ -67,17 +67,31 @@ export const featureFlagFilters = (features: Features | null): Array<FilterProps
...
+    FFM_7921_ARCHIVING_FEATURE_FLAGS: ARCHIVING_FEATURE_FLAGS
+  } = useFeatureFlags()
   let filters = featureFlagFilters(features)
 
   // remove 'Active Flags' filter card if feature flag FFM_3938_STALE_FLAGS_ACTIVE_CARD_HIDE_SHOW is not enabled
   if (!DISPLAY_ACTIVE_FILTER) {
     filters = filters.filter(filter => filter.label !== 'cf.flagFilters.active')
   }
 
+  // remove Archived filter card if there are no flags that are archived and if Archiving FF is toggled OFF
```
knagurski, line 91: Super minor, totally not a blocker
```suggestion
  // remove Archived filter card if there are no flags that are archived or if Archiving FF is toggled OFF
```

src/modules/75-cf/components/FlagOptionsMenuButton/__tests__/FlagOptionsMenuButton.test.tsx
```diff
@@ -331,7 +337,7 @@ describe('FlagOptionsButton', () => {
   describe('Archiving instead of Delete', () => {
     const isArchivingFfOn = true
```
knagurski, line 338: Not a blocker in the least, but since `FF` is short for `Feature Flag`, not Featureflag`, then both f's should be capitalised. #justsayin 😛 
```suggestion
    const isArchivingFFOn = true
```

src/modules/75-cf/components/FlagOptionsMenuButton/__tests__/FlagOptionsMenuButton.test.tsx
```diff
@@ -41,11 +42,12 @@ const renderComponent = (isArchivingFfOn = false, props: Partial<FlagOptionsMenu
   )
 }
 
-describe('FlagOptionsButton', () => {
+describe('FlagOptionsMenuButton', () => {
```
knagurski, line 45: 🙌🏻

src/modules/75-cf/components/FlagArchiving/useRestoreFlagDialog.tsx
```diff
@@ -0,0 +1,59 @@
...
+  return useConfirmAction({
+    title: getString('cf.featureFlags.archiving.restoreFlag'),
+    intent: Intent.DANGER,
+    confirmText: getString('cf.featureFlags.archiving.restore'),
+    message: (
+      <String stringID="cf.featureFlags.archiving.restoreDescription" vars={{ flagName: flagData.name }} useRichText />
+    ),
+    action: async () => {
+      handleRestore()
+    }
```
knagurski, line 55: Just curious if this can be simplified?
```suggestion
    action: handleRestore
```

src/modules/75-cf/components/FlagArchiving/useRestoreFlagDialog.tsx
```diff
@@ -0,0 +1,59 @@
...
+      showError(e.message)
+    }
+  }
+
+  return useConfirmAction({
+    title: getString('cf.featureFlags.archiving.restoreFlag'),
+    intent: Intent.DANGER,
+    confirmText: getString('cf.featureFlags.archiving.restore'),
+    message: (
+      <String stringID="cf.featureFlags.archiving.restoreDescription" vars={{ flagName: flagData.name }} useRichText />
```
knagurski, line 51: Looks like the actual string doesn't have any HTML in it. You can remove `useRichText`, otherwise `flagData.name` becomes a stored XSS attack vector.

src/modules/75-cf/components/FlagArchiving/useRestoreFlagDialog.tsx
```diff
@@ -0,0 +1,59 @@
...
+    ),
```
knagurski, line 52: I think the `message` in the confirmation dialog has a bit of stupid CSS attached to get text to break if it's too long, but we poorly done and causes words to get cut off in weird ways. We might need to add a wrapper around the content and apply the correct CSS.
```suggestion
    message: (
      <span style={{ wordBreak: 'break-word' }}>
        {getString('cf.featureFlags.archiving.restoreDescription', { flagName: flagData.name })}
      </span>
    ),
```

src/modules/75-cf/components/FlagArchiving/useRestoreFlagDialog.tsx
```diff
@@ -0,0 +1,59 @@
...
+  const { mutate: restoreFeatureFlag } = useRestoreFeatureFlag({
+    identifier: flagData.identifier,
+    queryParams
+  })
+
+  const handleRestore = async (): Promise<void> => {
+    try {
+      await restoreFeatureFlag()
+      showSuccess(getString('cf.featureFlags.archiving.restoreSuccess'))
+      refetchFlags?.()
```
knagurski, line 39: You can get rid of having to do this by defaulting `refetchFlags` to `() => void 0`. `() => {}` would also work, but ESLint will whinge at you
```suggestion
      refetchFlags()
```

src/modules/75-cf/components/FlagOptionsMenuButton/FlagOptionsMenuButton.tsx
```diff
@@ -92,36 +99,82 @@ const FlagOptionsMenuButton: FC<FlagOptionsMenuButtonProps> = ({
...
-      {
-        icon: (FFM_7921_ARCHIVING_FEATURE_FLAGS ? 'archive' : 'trash') as IconName,
-        text: FFM_7921_ARCHIVING_FEATURE_FLAGS ? getString('archive') : getString('delete'),
-        onClick: FFM_7921_ARCHIVING_FEATURE_FLAGS ? openDialog : confirmDeleteFlag,
-        permission: {
-          resource: { resourceType: ResourceType.FEATUREFLAG },
-          permission: PermissionIdentifier.DELETE_FF_FEATUREFLAG
+    let optionItems = [] as RbacMenuItemProps[]
+    if (FFM_7921_ARCHIVING_FEATURE_FLAGS) {
+      optionItems = [
```
knagurski, line 104: Super minor, but instead of having an array that's never used, but overwritten in a bunch of places, you can just return the new option items

src/modules/75-cf/components/FlagOptionsMenuButton/FlagOptionsMenuButton.tsx
```diff
@@ -92,36 +99,82 @@ const FlagOptionsMenuButton: FC<FlagOptionsMenuButtonProps> = ({
...
-        ...planEnforcementProps
-      },
-      {
-        icon: (FFM_7921_ARCHIVING_FEATURE_FLAGS ? 'archive' : 'trash') as IconName,
-        text: FFM_7921_ARCHIVING_FEATURE_FLAGS ? getString('archive') : getString('delete'),
-        onClick: FFM_7921_ARCHIVING_FEATURE_FLAGS ? openDialog : confirmDeleteFlag,
-        permission: {
-          resource: { resourceType: ResourceType.FEATUREFLAG },
-          permission: PermissionIdentifier.DELETE_FF_FEATUREFLAG
+    let optionItems = [] as RbacMenuItemProps[]
```
knagurski, line 102: If you do want to keep this, then type the variable, not the value
```suggestion
    let optionItems: RbacMenuItemProps[] = []
```

src/modules/75-cf/components/FlagOptionsMenuButton/FlagOptionsMenuButton.tsx
```diff
@@ -92,36 +99,82 @@ const FlagOptionsMenuButton: FC<FlagOptionsMenuButtonProps> = ({
...
+      optionItems = [
+        {
+          icon: flagData.archived ? 'redo' : 'edit',
+          text: flagData.archived ? getString('cf.featureFlags.archiving.restore') : getString('edit'),
+          onClick: flagData.archived ? openRestoreFlagDialog : gotoDetailPage,
+          permission: {
+            resource: { resourceType: ResourceType.ENVIRONMENT, resourceIdentifier: environment },
+            permission: PermissionIdentifier.EDIT_FF_FEATUREFLAG
+          },
+          ...planEnforcementProps
```
knagurski, line 113: I don't think we should combine these two buttons like this. They do fundamentally different things and the permissions are different.

src/modules/75-cf/components/FlagOptionsMenuButton/FlagOptionsMenuButton.tsx
```diff
@@ -92,36 +99,82 @@ const FlagOptionsMenuButton: FC<FlagOptionsMenuButtonProps> = ({
...
-        text: FFM_7921_ARCHIVING_FEATURE_FLAGS ? getString('archive') : getString('delete'),
-        onClick: FFM_7921_ARCHIVING_FEATURE_FLAGS ? openDialog : confirmDeleteFlag,
-        permission: {
-          resource: { resourceType: ResourceType.FEATUREFLAG },
-          permission: PermissionIdentifier.DELETE_FF_FEATUREFLAG
+    let optionItems = [] as RbacMenuItemProps[]
+    if (FFM_7921_ARCHIVING_FEATURE_FLAGS) {
+      optionItems = [
+        {
+          icon: flagData.archived ? 'redo' : 'edit',
```
knagurski, line 106: With all these ternaries, I think a better way would be to have an `if`/`else` instead.

src/modules/75-cf/components/FlagOptionsMenuButton/FlagOptionsMenuButton.tsx
```diff
@@ -92,36 +99,82 @@ const FlagOptionsMenuButton: FC<FlagOptionsMenuButtonProps> = ({
...
+          {
+            icon: 'trash',
+            text: getString('delete'),
+            onClick: confirmDeleteFlag,
+            permission: {
+              resource: { resourceType: ResourceType.FEATUREFLAG },
+              permission: PermissionIdentifier.DELETE_FF_FEATUREFLAG
+            },
+            ...planEnforcementProps
+          }
```
knagurski, line 162: A lot of the options are repeated. It might make sense to compose them instead of duplicating them. e.g.

```typescript
const options: Record<string, RbacMenuItemProps> = {
  restore: {
    // ...
  },
  edit: {
    // ...
  },
  archive: {
    // ...
  },
  delete: {
    // ...
  }
}

// ...

if (FFM_7921_ARCHIVING_FEATURE_FLAGS) {
  if (flagData.archived) {
    return [options.restore, options.delete]
  } else {
    return [options.edit, options.archive]
  }
}
// ...
```

crystal-fong: that works very nicely 👍

src/modules/75-cf/components/FlagOptionsMenuButton/FlagOptionsMenuButton.tsx
crystal-fong: I had to add that array bc TS was giving off that the contents of the array couldn't be applied to `RbacMenuItemProps`

src/modules/75-cf/components/FlagArchiving/useRestoreFlagDialog.tsx
crystal-fong: can do

src/modules/75-cf/pages/feature-flags/components/__tests__/FlagTableFilters.test.tsx
```diff
@@ -12,20 +12,27 @@ import { TestWrapper } from '@common/utils/testUtils'
 import mockFeatureFlags from '../../__tests__/mockFeatureFlags'
 import { FlagTableFilters, FlagTableFiltersProps } from '../FlagTableFilters'
 
-const updateTableFilter = jest.fn()
-let flagEnabled = true
-
-const renderComponent = (props?: Partial<FlagTableFiltersProps>): RenderResult =>
+// let flagEnabled = true
+// let archivingFFEnabled = false
```
d1wilko, line 16: probs worth removing these?

src/modules/75-cf/pages/feature-flags/components/__tests__/FlagTableFilters.test.tsx
```diff
@@ -39,12 +46,16 @@ const permanentFlagsFilter = {
 }
 
 describe('FlagTableFilters', () => {
-  beforeEach(() => {
-    flagEnabled = true
-  })
+  // beforeEach(() => {
+  //   flagEnabled = true
+  //   archivingFFEnabled = false
+  // })
```
d1wilko, line 52: worth removing this?

src/modules/75-cf/components/FlagOptionsMenuButton/FlagOptionsMenuButton.tsx
```diff
@@ -91,37 +98,70 @@ const FlagOptionsMenuButton: FC<FlagOptionsMenuButtonProps> = ({
...
+      permission: {
+        resource: { resourceType: ResourceType.ENVIRONMENT, resourceIdentifier: environment },
+        permission: PermissionIdentifier.EDIT_FF_FEATUREFLAG
+      },
+      ...planEnforcementProps
     }
+  }
 
-    return opts
+  const menuItems = useMemo(() => {
```
d1wilko, line 144: I know this isn't your original code, so I don't mind if you keep it as is but this menuItems function hurts my soul lol

I'd have it something like below - renamed so it is clear it is a function, no external variables so it is a pure function and get rid of useMemo because it is a premature optimisation on a really simple function - but I know it is not your code :)

`
const getMenuItems = (archiveFeatureFlag, flagData, options) => {
  if (archiveFeatureFlag) {
      if (flagData.archived) {
        return [options.restore, options.delete]
      } else {
        return [options.edit, options.archive]
      }
    } else {
      return [options.edit, options.delete]
    }
}
`

d1wilko (approved):
some small comments - but not worth slowing down the PR for :) good work!

Also should check if test failures are associated with this PR

src/modules/75-cf/pages/feature-flags/components/__tests__/FlagTableFilters.test.tsx
crystal-fong: 🤦‍♀️ yes

src/modules/75-cf/components/FlagArchiving/__tests__/useRestoreFlagDialog.test.tsx
```diff
@@ -0,0 +1,94 @@
...
+          orgIdentifier: 'mockOrgIdentifier',
+          projectIdentifier: 'mockProjectIdentifier'
+        }}
+        {...props}
+      />
+    </TestWrapper>
+  )
+
+  return result
+}
```
knagurski, line 39: No real need to assign to a variable and then return that variable
```suggestion
const renderComponent = (props: Partial<RestoreFlagDialogProps> = {}): RenderResult =>
  render(
    <TestWrapper>
      <WrapperComponent
        flagData={archivedMockFeature as Feature}
        onRefetch={jest.fn()}
        queryParams={{
          accountIdentifier: 'mockAccountIdentifier',
          orgIdentifier: 'mockOrgIdentifier',
          projectIdentifier: 'mockProjectIdentifier'
        }}
        {...props}
      />
    </TestWrapper>
  )
```

crystal-fong: this doesn't seem to be a blocker but I can address it in the next PR

src/modules/75-cf/components/FlagArchiving/useRestoreFlagDialog.tsx
```diff
@@ -0,0 +1,60 @@
...
+  })
+
+  const handleRestore = async (): Promise<void> => {
+    try {
+      await restoreFeatureFlag()
+      showSuccess(getString('cf.featureFlags.archiving.restoreSuccess'))
+      onRefetch()
+    } catch (e) {
+      handleResponseError(e)
+      showError(e.message)
```
knagurski, line 43: The reason you're having to use `getAllByText` in the test for this is because `handleResponseError` is doing this exact thing (`src/modules/75-cf/hooks/useResponseError.tsx` line 33) and in fact the app is displaying double errors. You can delete this line and it should work fine.

src/modules/75-cf/components/FlagArchiving/useRestoreFlagDialog.tsx
```diff
@@ -0,0 +1,60 @@
...
+    }
+  }
+
+  return useConfirmAction({
+    title: getString('cf.featureFlags.archiving.restoreFlag'),
+    intent: Intent.DANGER,
+    confirmText: getString('cf.featureFlags.archiving.restore'),
+    message: (
+      <span className={css.restoreMessage}>
+        <String stringID="cf.featureFlags.archiving.restoreDescription" vars={{ flagName: flagData.name }} />
```
knagurski, line 53: Out of curiosity, why use the `<String />` component instead of just `getString`?
```suggestion
        {getString('cf.featureFlags.archiving.restoreDescription', { flagName: flagData.name })}
```

src/modules/75-cf/components/FlagArchiving/useRestoreFlagDialog.tsx
```diff
@@ -0,0 +1,60 @@
...
+import { String, useStrings } from 'framework/strings'
+import { useRestoreFeatureFlag, RestoreFeatureFlagQueryParams, Feature } from 'services/cf'
+import useResponseError from '@cf/hooks/useResponseError'
+import { useConfirmAction } from '@common/hooks'
+import css from './useRestoreFlagDialog.module.scss'
+
+export interface RestoreFlagDialogProps {
+  flagData: Feature
+  queryParams: RestoreFeatureFlagQueryParams
+  onRefetch: () => void
```
knagurski, line 13: Not a blocker, but this should be probably be `onRestore`, not `onRefetch` since this component isn't refetching and then calling this, it's restoring and then calling it.

src/modules/75-cf/components/FlagOptionsMenuButton/FlagOptionsMenuButton.tsx
```diff
@@ -91,39 +98,68 @@ const FlagOptionsMenuButton: FC<FlagOptionsMenuButtonProps> = ({
...
+      ...planEnforcementProps
+    },
+    restore: {
+      icon: 'redo',
+      text: getString('cf.featureFlags.archiving.restore'),
+      onClick: openRestoreFlagDialog,
+      permission: {
+        resource: { resourceType: ResourceType.ENVIRONMENT, resourceIdentifier: environment },
+        permission: PermissionIdentifier.DELETE_FF_FEATUREFLAG
+      },
```
knagurski, line 129: This should have the same permissions as archive.

src/modules/75-cf/components/FlagArchiving/useRestoreFlagDialog.tsx
crystal-fong: just habit since it what I used whenever I needed a string with a var

src/modules/75-cf/components/FlagArchiving/useRestoreFlagDialog.tsx
crystal-fong: ??? don't understand. It's the refetch flag function I'm passing in as prop

src/modules/75-cf/components/FlagOptionsMenuButton/FlagOptionsMenuButton.tsx
knagurski: Archive/Delete:
```
permission: {
  resource: { resourceType: ResourceType.FEATUREFLAG },
  permission: PermissionIdentifier.DELETE_FF_FEATUREFLAG
}
```

Restore:
```
permission: {
  resource: { resourceType: ResourceType.ENVIRONMENT, resourceIdentifier: environment },
  permission: PermissionIdentifier.DELETE_FF_FEATUREFLAG
}
```

They ain't the same

crystal-fong: 👍

knagurski (approved):
Really awesome work! Shaping up very nicely!

laura-mannus (approved):

PR #16512, opened 2023-06-22
fix: [CDS-72807]: fix an issue causing yaml to not update when clearing fields

rameshx:
Cherry-pick of https://github.com/harness/harness-core-ui/pull/16508/commits


PratyushGar (approved):

src/modules/70-pipeline/components/PipelineSteps/AdvancedSteps/AdvancedSteps.tsx
```diff
@@ -227,7 +227,7 @@ export function AdvancedTabForm(props: AdvancedTabFormProps): React.ReactElement
                             if (isMultiTypeRuntime(type)) {
                               set(draft, 'when', RUNTIME_INPUT_VALUE)
                             } else {
-                              unset(draft, 'when')
+                              set(draft, 'when', undefined)
```
AshwinBhatkal-Harness, line 230: Is it not the same as unset? 🤔

rameshx: `unset` is like `delete object.key`, I want to do `object[key] = undefined`

PR #16636, opened 2023-06-28
feat: [CDS-72672]: Add repo allowlist setting under GitX, send query param to filter repos 

rameshx:
### Summary

- This adds a new setting Repo Allowlist under GitX
- Sends a new query param to filter repositories based on the setting

#### Screenshots

https://github.com/harness/harness-core-ui/assets/52284933/ef074bb9-efbd-483e-ba3f-8bfe39581dda


src/modules/23-default-settings/components/ReusableHandlers.tsx
```diff
@@ -191,3 +192,11 @@ export const DefaultSettingsToggle: React.FC<SettingRendererProps> = ({
     </Layout.Vertical>
   )
 }
+
+export interface DefaultSettingsTagInputProps extends SettingRendererProps {
+  tagInputProps?: Omit<KVTagInputProps, 'name'>
+}
+
+export const DefaultSettingsTagInput = ({ identifier, tagInputProps }: DefaultSettingsTagInputProps): JSX.Element => {
```
sunnykesh-harness, line 200: Does `KVTagInput` fits here ? it may be a combination of `key:value` as well. What we need here is just a value here.

src/modules/23-default-settings/components/ReusableHandlers.tsx
```diff
@@ -191,3 +192,11 @@ export const DefaultSettingsToggle: React.FC<SettingRendererProps> = ({
...
+  return <FormInput.KVTagInput className={css.defaultSettingRenderer} name={identifier} {...tagInputProps} />
```
sunnykesh-harness, line 201: Does KVTagInput fits here ? it may be a combination of key:value as well. What we need here is just a value here.

rameshx: `KVTagInput` can be customized to have `key:value` or just `value` by modifying `tagInputProps`. For the GitX setting that I added, `key:value` is not needed, but `DefaultSettingsTagInput` should still use `KVTagInput` as this should be a reusable handler.

sunnykesh-harness (approved):

samarth-sharma31 (dismissed):

ishantmehtaharness (dismissed):
Looks Good Ramesh

ishantmehtaharness (approved):
looks good

PR #16749, opened 2023-07-05
fix: [FFM-8576]: Ensure flags and environments load when not runtime

knagurski:
This PR fixes an issue where flags and environments do not load when they are set as fixed instead of runtime. It also fixes an issue where a call to get a specific flag would fail on a loop and another issue where the form would fail to load in template view.


src/modules/75-cf/components/PipelineSteps/FlagConfigurationStep/FlagConfigurationStepWidget.tsx
```diff
@@ -216,7 +246,10 @@ const FlagConfigurationStepWidget = forwardRef(
         })}
       >
         {formik => {
-          setFormikRef(formikRef, formik)
+          if (formikRef && 'current' in formikRef) {
+            setFormikRef(formikRef, formik)
+          }
```
knagurski, line 251: Had to conditionally do this coz in some views it passes an empty object as `formikRef` instead of an actual ref. Weird.

Harness-JKhattak (approved):
🚀

laura-mannus (approved):

src/modules/75-cf/components/PipelineSteps/FlagConfigurationStep/FlagConfigurationStepWidget.tsx
bgribben-harness: Would it be worthwhile to have  `if (formikRef && formikRef.current)` rather than using `in`? Just thinking of a scenario where the key is present but is null/undefined.

Also, I'm not familiar with the `in` keyword so maybe that scenario is already taken care of.

knagurski: I had wanted to do it that way, but in the initial render when `current` is set, it's set as `null` and that's totally valid. We just need to make sure it exists, not check it as being truthy

bgribben-harness: Ah sweet

bgribben-harness (approved):

ishantmehtaharness (approved):
Looks Good Kevin

PR #16942, opened 2023-07-18
feat: [FFM-8656]: dependent flags

crystal-fong:
### Summary
This PR shows a modal where a user attempts to archive a flag that has dependents. If a flag doesn't have dependents, it will show a modal that allows a flag to be archived 

#### Screenshots

https://github.com/harness/harness-core-ui/assets/95680496/471f9977-f3de-4db4-8db0-a3419bfd8a4c

#### PR Checklist

<details>

</details>

<details>

- Jest: `retrigger jest`
- ESLint: `retrigger eslint`
- Sonar: `retrigger sonar`
- Build: `retrigger build`
- Rebase: `trigger rebase`
</details>


src/modules/75-cf/components/FlagOptionsMenuButton/FlagOptionsMenuButton.tsx
```diff
@@ -112,11 +116,24 @@ const FlagOptionsMenuButton: FC<FlagOptionsMenuButtonProps> = ({
     )
   }
 
+  const onClickArchive = useCallback(() => {
+    getDependentFeaturesPromise({ identifier: flagData.identifier, queryParams }).then(result => {
```
crystal-fong, line 120: unsure whether to use this since I only want to call this endpoint whenever I click on archive button. `useGetDependentFeatures` seems to fetch the result for each flag which isn't necessary

src/modules/75-cf/components/FlagArchiving/useArchiveFlagDialog.tsx
```diff
@@ -26,14 +37,64 @@ const useArchiveFlagDialog = ({
...
+  // removes the flagFilter queryParam which is undefined
+  const dependentFlagQueryParams = useMemo(
+    () => Object.fromEntries(Object.entries(queryParams).filter(([_, v]) => v != null)),
+    [queryParams]
+  )
+
+  const state = useMemo<STATUS>(() => {
+    if (dependentFlagsError) {
+      return STATUS.error
+    } else if (dependentFlagsLoading) {
```
laura-mannus, line 79: should these be just if's?

src/modules/75-cf/components/FlagArchiving/useArchiveFlagDialog.tsx
```diff
@@ -46,36 +107,65 @@ const useArchiveFlagDialog = ({
     }
   }
 
+  const getDialogContent = () => {
+    if (state === STATUS.error)
```
laura-mannus, line 111: NBD: This might be a good place to use a switch than if's

src/modules/75-cf/components/FlagDetailsOptionsMenuButton/FlagDetailsOptionsMenuButton.tsx
```diff
@@ -118,7 +125,7 @@ const FlagDetailsOptionsMenuButton: FC<FlagDetailsOptionsMenuButtonProps> = ({
     archive: {
       icon: 'archive',
       text: getString('archive'),
-      onClick: openArchiveDialog,
+      onClick: onClickArchiveButton,
```
laura-mannus, line 128: Just curious, is it better to extract into a const? Whats the reasoning for not doing this inline?

crystal-fong: just personal preference and neatness

src/modules/75-cf/components/FlagDetailsOptionsMenuButton/__tests__/FlagDetailsOptionsMenuButton.test.tsx
```diff
@@ -327,6 +328,43 @@ describe('FlagDetailsOptionsMenuButton', () => {
...
+        loading: false,
+        refetch: jest.fn()
+      } as any)
+
+      renderComponent(isArchivingFFOn)
+
+      await userEvent.click(document.querySelector('[data-icon="Options"]') as HTMLButtonElement)
+      await userEvent.click(document.querySelector('[data-icon="archive"]') as HTMLButtonElement)
+
+      expect(screen.getAllByText(error)[0]).toBeInTheDocument()
```
laura-mannus, line 366: Is multiple errors expected? :O

laura-mannus: 😭

laura-mannus (approved):
Comments are not blockers, all looks really good imo well done this wasn't easy

src/modules/75-cf/components/FlagDetailsOptionsMenuButton/__tests__/FlagDetailsOptionsMenuButton.test.tsx
crystal-fong: I know it sucks, but hopefully when this (eventually) gets migrated to the MFE, theres a line that I wrote that does remove the toasters between every test in the whole repo, not just per test suite :D

src/modules/75-cf/components/FlagDetailsOptionsMenuButton/FlagDetailsOptionsMenuButton.tsx
knagurski: Ideally it should be passed through `useCallback`, otherwise passing an inline function _could_ trigger unnecessary rerenders, or even in the worst case an infinite loop as the reference changes on every render.

src/modules/75-cf/components/FlagDetailsOptionsMenuButton/__tests__/FlagDetailsOptionsMenuButton.test.tsx
knagurski: Actually if you run the test in isolation it does produce 2, so not just leakage from other tests. Looks like there's a bug in there.

src/modules/75-cf/components/FlagArchiving/CannotArchiveWarning.tsx
```diff
@@ -6,50 +6,77 @@
...
-                  projectIdentifier: projectIdentifier as string,
-                  featureFlagIdentifier: flag.feature,
-                  accountId
-                })
-              )}
-            >
-              {flag.feature}
-            </Link>
+            <Text>{flag.identifier}</Text>
           </Layout.Horizontal>
```
knagurski, line 64: This is perhaps a wee bit more complex than needed.
```typescript
<Text>{getString('common.ID')}: {flag.identifier}</Text>
```
That should look the same, but get rid of the `Layout` and an extra `Text`

src/modules/75-cf/components/FlagArchiving/CannotArchiveWarning.tsx
```diff
@@ -6,50 +6,77 @@
...
         </ItemContainer>
       ))}
-      <Text font={{ variation: FontVariation.BODY2 }}>{getString('cf.featureFlags.archiving.removeFlag')}</Text>
+      <Text padding={{ top: 'medium' }} color={Color.GREY_500} font={{ variation: FontVariation.BODY2 }}>
+        {getString('cf.featureFlags.archiving.removeFlag')}
+      </Text>
+      <Pagination
+        gotoPage={pageNum => {
+          setPageNumber(pageNum)
+        }}
```
knagurski, line 73: Super-minor and deffo not a blocker, but you can simplify here
```suggestion
        gotoPage={setPageNumber}
```

src/modules/75-cf/components/FlagArchiving/useArchiveFlagDialog.tsx
```diff
@@ -46,36 +107,65 @@ const useArchiveFlagDialog = ({
...
-          onChange={e => {
-            setValidationFlagIdentifier((e?.target as HTMLInputElement)?.value)
-            setIsAnIdentifierMatch((e?.target as HTMLInputElement)?.value === flagData?.identifier)
-          }}
-        />
-      </Layout.Vertical>
-    ),
-    customButtons: (
+    contentText: getDialogContent(),
+    customButtons: state === STATUS.hasNoDependentFlags && (
```
knagurski, line 168: At this point you're not getting any of the benefits of using the confirmation dialog. It would probably be easier to just use a `ModalDialog` component with `useModalHook`. You'd then be able to drop the CSS overrides and have a much better visual experience.

src/modules/75-cf/components/FlagArchiving/useArchiveFlagDialog.tsx
```diff
@@ -46,36 +107,65 @@ const useArchiveFlagDialog = ({
     }
   }
 
+  const getDialogContent = () => {
```
knagurski, line 110: Currently this gets re-evaluated and contents rerendered on each rerender of the the dialog. A better idea would be to put it behind a `useMemo` to stabilise the references.
```suggestion
  const dialogContent = useMemo(() => {
```

src/modules/75-cf/components/FlagArchiving/useArchiveFlagDialog.tsx
```diff
@@ -46,36 +107,65 @@ const useArchiveFlagDialog = ({
...
+    if (state === STATUS.error)
+      return <PageError message={showError(dependentFlagsError)} onClick={() => refetchDependentFlags()} />
```
knagurski, line 112: Try to avoid multiline things like this without brackets. A quick visual scan would make you think that the next statements were inside this `if` block.
```suggestion
    if (state === STATUS.error) {
      return <PageError message={showError(dependentFlagsError)} onClick={() => refetchDependentFlags()} />
    }
```

src/modules/75-cf/components/FlagArchiving/useArchiveFlagDialog.tsx
```diff
@@ -46,36 +107,65 @@ const useArchiveFlagDialog = ({
...
+          flagIdentifier={flagData.identifier}
+          flagName={flagData.name}
+          dependentFlagsResponse={dependentFlagsResponse as Features}
+          queryParams={dependentFlagQueryParams as GetDependentFeaturesQueryParams}
+          pageNumber={pageNumber}
+          setPageNumber={setPageNumber}
+        />
+      )
+    }
+    if (state === STATUS.hasNoDependentFlags) {
```
knagurski, line 127: Make sure to use line breaks around conditions like this to break up the wall of code and make it visually scannable.
```suggestion
    if (state === STATUS.loading) return <ContainerSpinner flex={{ align: 'center-center' }} />
    
    if (state === STATUS.hasDependentFlags) {
      return (
        <CannotArchiveWarning
          flagIdentifier={flagData.identifier}
          flagName={flagData.name}
          dependentFlagsResponse={dependentFlagsResponse as Features}
          queryParams={dependentFlagQueryParams as GetDependentFeaturesQueryParams}
          pageNumber={pageNumber}
          setPageNumber={setPageNumber}
        />
      )
    }
    
    if (state === STATUS.hasNoDependentFlags) {
```

src/modules/75-cf/components/FlagArchiving/useArchiveFlagDialog.tsx
knagurski, line 112: `message` is expected to be what's shown to the user. `showError` is a function that shows the toaster. They're not really compatible. Either we should show the error using `PageError` or we should show the toaster, not both. In this case, since it's the fetch and display, I'd just show the error and ditch the toaster.
```suggestion
      return <PageError message={dependentFlagsError} onClick={() => refetchDependentFlags()} />
```

src/modules/75-cf/components/FlagArchiving/CannotArchiveWarning.tsx
```diff
@@ -6,50 +6,77 @@
...
+        itemCount={dependentFlagsResponse.itemCount || 0}
+        pageCount={dependentFlagsResponse.pageCount || 0}
+        pageIndex={pageNumber || 0}
```
knagurski, line 76: Looks like this is actually always going to be a number, so you can drop the `|| 0`
```suggestion
        pageIndex={pageNumber}
```

src/modules/75-cf/components/FlagArchiving/useArchiveFlagDialog.tsx
```diff
@@ -46,36 +107,65 @@ const useArchiveFlagDialog = ({
...
+                : undefined
+            }
+            intent={!isAnIdentifierMatch && validationFlagIdentifier !== '' ? Intent.DANGER : Intent.PRIMARY}
+            aria-label={getString('cf.featureFlags.archiving.confirmFlag')}
+            placeholder={flagData?.identifier}
+            value={validationFlagIdentifier}
+            onChange={e => {
+              setValidationFlagIdentifier((e?.target as HTMLInputElement)?.value)
+              setIsAnIdentifierMatch((e?.target as HTMLInputElement)?.value === flagData?.identifier)
+            }}
```
knagurski, line 150: If you type `e` and use `currentTarget`, you can drop the type assertions.
```suggestion
            onChange={(e:FormEvent<HTMLInputElement>) => {
              setValidationFlagIdentifier(e.currentTarget.value)
              setIsAnIdentifierMatch(e.currentTarget.value === flagData?.identifier)
            }}
```

src/modules/75-cf/components/FlagDetailsOptionsMenuButton/FlagDetailsOptionsMenuButton.tsx
```diff
@@ -5,15 +5,15 @@
...
+import React, { FC, useState } from 'react'
 import { useHistory, useParams } from 'react-router-dom'
 import type { MutateMethod, MutateRequestOptions } from 'restful-react/dist/Mutate'
 import usePlanEnforcement from '@cf/hooks/usePlanEnforcement'
 import RbacOptionsMenuButton from '@rbac/components/RbacOptionsMenuButton/RbacOptionsMenuButton'
 import { FeatureIdentifier } from 'framework/featureStore/FeatureIdentifier'
 import { useStrings } from 'framework/strings'
 import { useFeatureFlags } from '@common/hooks/useFeatureFlag'
-import type {
+import {
```
knagurski, line 16: Any specific reason to remove the `type` from this? It appears that all the imported items are interfaces, not concretes.

src/modules/75-cf/components/FlagArchiving/useArchiveFlagDialog.tsx
crystal-fong: but we hate `useModalHook` too

crystal-fong: unless this is just a case of picking the lesser evil

knagurski: We do hate `useModalHook`, but `useConfirmationDialog` is just an old style dialog with 2 pre-made buttons and an icon inside a `useModalHook` anyway. So swapping it for a `ModalDialog` in a `useModalHook` will give you more control and less tradeoffs. And it'll be easier to refactor later to eliminate the hook.

src/modules/75-cf/components/FlagArchiving/useArchiveFlagDialog.tsx
Kira-Jennings: Or a switch

src/modules/75-cf/components/FlagArchiving/useArchiveFlagDialog.tsx
crystal-fong: neat! TY

src/modules/75-cf/components/FlagArchiving/CannotArchiveWarning.tsx
crystal-fong: that is true

crystal-fong: this works but I had to modify the test because the text rendered in the dom looks strange now a line break between `common.ID`, the colon and the flag identifier

src/modules/75-cf/components/FlagArchiving/useArchiveFlagDialog.tsx
```diff
@@ -26,14 +37,64 @@ const useArchiveFlagDialog = ({
...
+    'error',
+    'hasDependentFlags',
+    'hasNoDependentFlags',
+    'loading',
+    'ok'
+  }
+
+  // removes the flagFilter queryParam which is undefined
+  const dependentFlagQueryParams = useMemo(
+    () => Object.fromEntries(Object.entries(queryParams).filter(([_, v]) => v != null)),
```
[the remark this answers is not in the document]
crystal-fong: good spot

src/modules/75-cf/components/FlagArchiving/useArchiveFlagDialog.tsx
crystal-fong: this must have been causing the duplicate error messages appearing in the test suites

Kira-Jennings (approved):

src/modules/75-cf/components/FlagDetailsOptionsMenuButton/__tests__/FlagDetailsOptionsMenuButton.test.tsx
crystal-fong: this has been fixed now

src/modules/75-cf/components/FlagDetailsOptionsMenuButton/FlagDetailsOptionsMenuButton.tsx
laura-mannus: Hey, think we should address what Kev says here and use `useCallback`

crystal-fong: missed this, added now thanks

laura-mannus (approved):

Kira-Jennings (approved):

knagurski (approved):
Niiiiiiccccceeeee

PR #17049, opened 2023-07-24
fix: [PL-40144]: fix default page size in Audit Trail

rameshx:
### Summary

When the FF `PL_NEW_PAGE_SIZE` is disabled, the default page size of Audit Trial should be 20 instead of 25.

#### Screenshots

Before:
Screenshot 2023-07-24 at 4 20 45 PM

After:
Screenshot 2023-07-24 at 4 23 29 PM


src/modules/10-common/hooks/useQueryParams.ts
rameshx, line 1: Changes in this file are unrelated. Just added comments to make it easier to understand.

avinashmadhwani02 (approved):

pankajsingh-harness (approved):

PR #17186, opened 2023-07-31
fix: [FFM-8680]: Use in-memory type-ahead search for Target Groups on Flag detail page

knagurski:
This PR fixes an issue where using the type-ahead search in the Target Groups selectors will blank previously selected Target Groups. This is a temporary fix and should be revisited to change how Target Groups are fetched to ensure initial values and searches do not step on each others toes.
|Before|After|
|------|------|
|Before|After|


src/modules/75-cf/pages/feature-flags-detail/targeting-rules-tab/TargetingRulesTab.tsx
```diff
@@ -67,8 +67,8 @@ const TargetingRulesTab: FC<TargetingRulesTabProps> = ({ featureFlagData, refetc
     debounce
   })
 
-  const { data: segmentsData, refetch: refetchSegments } = useGetAllSegments({
-    queryParams: queryParams as GetAllSegmentsQueryParams,
+  const { data: segmentsData } = useGetAllSegments({
+    queryParams: { ...(queryParams as GetAllSegmentsQueryParams), pageSize: 500 },
```
knagurski, line 71: The max number of target groups we have currently in production is `103` in a test account, so `500` is just to give us _plenty_ of headroom.

bgribben-harness (approved):

Harness-JKhattak (approved):

PR #17224, opened 2023-08-01
fix: [CDS-75558]: fix instance list panel for child, rollback graphs

rameshx:
If pipeline stages are used, the execution graph of the nested stages can be accessed using `childGraph`. This updates the instance list panel to find and use the correct execution graph when Show All is clicked on a collapsed node.

https://github.com/harness/harness-core-ui/assets/52284933/87c052e9-6ddb-4c5f-94ad-e041f4dc4651


PratyushGar (approved):

rameshx:
This can be force merged as coverage for `StepGroupNode.tsx` is low. Jira for increasing the coverage: https://harness.atlassian.net/browse/CDS-72199

litvinovstas (approved):

ishantmehtaharness (approved):
looks good

rameshx:
@kalyansai-harness This is the original design: https://www.figma.com/file/Hudsz2Rdq5xQN5yL81bBGJ/SSH---Deployment-Swimlane?type=design&node-id=2159-256588&mode=design&t=pTnGeJNCzli6kNA7-0

As I explained, we had to ultimately remove the hamburger icon and instead make the name clickable because of the status info at the top right and we also decided to update the Show All button to display the same instance list as the list shows all nodes.

Related conversation with the designer:

> [RR]: Is the new hamburger icon at the top right meant for opening the list? currently the container shows success and fail counts at the top right
> [YC]: We can apply the same status counts instead of the hamburger. (and yup, hamburger -- list icon was supposed to open the panel too, but we can exclude it for now and just show status counts instead)

> [RR]: The container that renders the nodes has a 'Show all' button that shows all nodes when clicked, is there any difference between using this button and selecting all nodes in the new window?
> [YC]: it would be the same, since it should show all the list

PR #17271, opened 2023-08-03
fix: [BG-1189]: only show an error if there's a message in it

rameshx:
### Summary

Based on the non-minified error in https://app.bugsnag.com/harness-1/prod-app-dot-harness-dot-io/errors/64c890f51630f400086736fe?event_id=64c890f500bee5b2e49a0000&i=jr&m=ci, `showError` is getting called with an object instead of a string. This PR extracts the error message and sends it to `showError` to fix the issue.


PratyushGar (approved):

pankajsingh-harness (approved):

chetanvekariya-harness (approved):

src/modules/10-common/components/UserGroupsReference/UserGroupsReference.tsx
```diff
@@ -122,8 +122,12 @@ const fetchRecords = async (
       done(response)
     }
   } catch (error) {
-    showError(error as string)
     done([])
+
+    const errorMessage = getErrorInfoFromErrorObject(error)
```
chetanvekariya-harness, line 127: We can have this inside if condition.

chetanvekariya-harness (approved):

PR #17289, opened 2023-08-04
feat: [CODE-713]: Expose Gitness API service to harness-core-ui

tan-nhu:
### Summary

Expose Gitness API service to harness-core-ui. The service is required to implement Gitness/CI integration.

#### Screenshots

NA

#### PR Checklist

<details>

</details>

<details>

- Jest: `retrigger jest`
- ESLint: `retrigger eslint`
- Sonar: `retrigger sonar`
- Build: `retrigger build`
- Rebase: `trigger rebase`
</details>


tan-nhu:
retrigger titlecheck

shaurya-harness (dismissed):

configs/restful-react.config.js
```diff
@@ -240,6 +240,15 @@ module.exports = {
       base: `{getConfig("ccm/public-pricing/api/v1")}`
     }
   },
+  code: {
```
abhinavrastogi-harness, line 243: can we use the new openapi client instead?
once we add this, it'll take many quarters to get it removed again due to code dependencies.

src/services/code/overrides.yaml
```diff
@@ -0,0 +1,2 @@
+allowpaths: ['*']
```
abhinavrastogi-harness, line 1: why are all paths being added? we should allow only the used paths to optimise what we ship.

src/services/code/swagger.yaml
abhinavrastogi-harness, line 1: can we avoid committing the api spec? it bloats the repo unnecessarily.

abhinavrastogi-harness (changes requested):

configs/restful-react.config.js
tan-nhu: We are using restful-react in Gitness, eventually we'll migrate to the new openapi at some point, but it needs to come from Gitness first. Right now, I don't want to make it complicated for the teams to work on with two sets of API calls.

src/services/code/overrides.yaml
tan-nhu: Added allowPaths to `repos` only for now.

src/services/code/swagger.yaml
tan-nhu: I don't think we should remove the spec. It's there to keep track of changes from API side.

abhinavrastogi-harness: APi changes will reflect in the generated service automatically. 
You can add it to `.gitignore` if you still want to keep it in the repo.
Do you see any specific benefits from committing the API spec?

configs/restful-react.config.js
abhinavrastogi-harness: It'll still be the same API even if it's a different client. I don't think we are planning to pass these hooks down to the microfrontend, right? Gitness MFE  already has it's own clients generated.

When is Gitness planned to be migrated to OpenAPI?
Why is this service needed in NGUI in the first place?

src/services/code/swagger.yaml
tan-nhu: The spec is used as a reference what is changed from backend side (the source), not what is generated (the output). It's used to communicate as the source of truth from backend. Let's say if backend publishes some breaking changes, without the source, you can't tell and track back. 

Versioning is important.

configs/restful-react.config.js
tan-nhu: It's not the same API if you have naming different.

`When is Gitness planned to be migrated to OpenAPI?`

No date specified. I don't think Gitness is any different than other API services currently under resful-react.

`Why is this service needed in NGUI in the first place?`

It's for CIE integration. CIE needs a way to call Gitness APIs directly.

hitesharinga:
Approving since CI team is waiting for this.

PR #17351, opened 2023-08-10
fix: [CDS-76892]: use empty string when value/original is falsy to avoid page crash

rameshx:
Update `MonacoDiffEditor` to use empty strings for `value` and `original` props if they are falsy. The types exposed by the library are incorrect as `value`, `original` are typed as `string | undefined`, but passing `undefined` throws an error.


src/modules/75-cf/components/AuditLogs/EventSummary.tsx
```diff
@@ -180,24 +180,6 @@ export const EventSummary: React.FC<EventSummaryProps> = ({ data, flagData, onCl
...
-                          startColumn: 0,
-                          endLineNumber: 0,
-                          endColumn: 0,
-                          selectionStartLineNumber: 0,
-                          selectionStartColumn: 0,
-                          positionLineNumber: 0,
-                          positionColumn: 0
-                        })
-                      }, 0)
-                    }}
```
rameshx, line 200: This is probably added back then to remove any code selection on mount. Removed this as it's not required now.

crystal-fong (approved):

pankajsingh-harness (approved):

PR #17535, opened 2023-08-23
fix: [FFM-9033]: Accommodate rules with multiple target groups

knagurski:
This PR fixes an issue where rules created with the `AddSegmentToVariationTargetMap` instruction with multiple Target Groups would only display the 1st Target Group. This PR detects when rules have been defined in this way when changed and then breaks them apart into different rules.

|Before|After|
|------|------|
|Before|After|


src/modules/75-cf/pages/feature-flags-detail/targeting-rules-tab/hooks/useTargetingRulesFormData.tsx
```diff
@@ -49,15 +49,20 @@ const useTargetingRulesFormData = ({
...
-      featureFlagData.envProperties?.rules
-        ?.filter(rule => rule.serve.variation === variation.identifier)
-        .map(targetGroupRule => ({
-          priority: targetGroupRule.priority,
-          label: segments.find(segment => segment.identifier === targetGroupRule.clauses[0].values[0])?.name as string,
-          ruleId: targetGroupRule.ruleId as string,
-          value: targetGroupRule.clauses[0].values[0]
-        })) || []
+    const variationTargetGroups: VariationTargetGroup[] = (featureFlagData.envProperties?.rules || [])
+      .filter(rule => rule.serve.variation === variation.identifier)
```
bgribben-harness, line 53: Could the check in this filter not be made in the reduce function?

src/modules/75-cf/pages/feature-flags-detail/targeting-rules-tab/utils/PatchFeatureFlagUtils.ts
```diff
@@ -84,8 +87,51 @@ export const PatchFeatureFlagUtils = (
...
       submittedVariations.find(x => x.variationIdentifier === variationIdentifier)?.targetGroups || []
 
-    return initialTargetGroups.filter(
-      targetGroup => !submittedTargetGroups.map(({ label }) => label).includes(targetGroup.label)
+    return (
+      initialTargetGroups
+        // get groups that aren't in the form submission
+        .filter(targetGroup => !submittedTargetGroups.map(({ value }) => value).includes(targetGroup.value))
+        // exclude groups that share a ruleId with a non-removed group
+        .filter(
```
bgribben-harness, line 95: Could this not be condensed into a single filter call?

src/modules/75-cf/pages/feature-flags-detail/targeting-rules-tab/hooks/useTargetingRulesFormData.tsx
knagurski: It absolutely could, but by doing it separately, it's a bit easier to reason about and follow the flow. Each operation has a single task, gradually manipulating things as it goes.

src/modules/75-cf/pages/feature-flags-detail/targeting-rules-tab/utils/PatchFeatureFlagUtils.ts
```diff
@@ -84,8 +87,51 @@ export const PatchFeatureFlagUtils = (
...
+                .map(({ value }) => value)
+            }
+
+            updatedRules.push(currentRule)
+          }
+
+          currentRule.values = currentRule.values.filter(targetGroupId => targetGroupId !== targetGroup.value)
+
+          return updatedRules
+        }, [])
```
bgribben-harness, line 134: Again, why not put the filter checks in the reduce? Readability?

src/modules/75-cf/pages/feature-flags-detail/targeting-rules-tab/utils/PatchFeatureFlagUtils.ts
knagurski: Yup, but I find it useful to gradually do these things in different operations so that I can reason about them better. Doing them in the same operation would be fine, but I find it more prone to mistakes in logic.

src/modules/75-cf/pages/feature-flags-detail/targeting-rules-tab/utils/PatchFeatureFlagUtils.ts
knagurski: Exactly. The code in this file absolutely fries my head and separating out into concrete steps that I can see makes it easier to follow and reason about.

laura-mannus (approved):

src/modules/75-cf/pages/feature-flags-detail/targeting-rules-tab/utils/PatchFeatureFlagUtils.ts
bgribben-harness: haha aye I expanded the code around the changes a bit and it got very overwhelming very quickly!

Are there any performance concerns around chaining these calls or is the typical dataset pretty slim? I'm sure wherever they're being called is all useMemo'd up 👀

willd0yle (approved):

src/modules/75-cf/pages/feature-flags-detail/targeting-rules-tab/utils/PatchFeatureFlagUtils.ts
knagurski: Great question. The data sets are pretty small and not expected to be very large (we're talking single or double digits, not millions). The good thing about repeated `filters` is that the size of the dataset is slimming down on each; depending on the data and how the filters are written, this can actually be a huge perf win over doing things in a single chunk. Not in this case though, I'd wager that that perf difference between for this case is sub-millisecond; but the readability and reasonability is much greater IMO.

bgribben-harness (approved):

crystal-fong (approved):

PR #17748, opened 2023-09-06
feat: [FFM-8350]: Changes changes v2 and Set Flag Switch

knagurski:
This PR adds a v2 of the Flag Changes section of the FF pipeline step. When it goes GA I will kill off the v2 and remove the older version.

|Before|After|
|------|------|
|Before|After|


src/modules/75-cf/components/PipelineSteps/FlagConfigurationStep/FlagChangesV2/FlagChangesForm.tsx
```diff
@@ -0,0 +1,118 @@
...
+import { FeatureFlagConfigurationInstruction, FlagConfigurationStepFormDataValues } from '../types'
+import { SubSectionComponent, subSectionNames } from './subSection.types'
+import SubSections from './SubSections'
+
+// sub-sections
+import SetFlagSwitch from './subSections/SetFlagSwitch'
+
+export const allSubSections: SubSectionComponent[] = [
+  SetFlagSwitch
+  // DefaultRules,
```
crystal-fong, line 24: are all the commented out code TODOs?

knagurski: Yup 👍🏻 It's big and complex and I wanted to keep some hints in there while it's still bing worked on behind a flag. My brain hurts

src/modules/75-cf/components/PipelineSteps/FlagConfigurationStep/FlagChangesV2/FlagChanges.module.scss
```diff
@@ -0,0 +1,21 @@
+/*
+ * Copyright 2023 Harness Inc. All rights reserved.
+ * Use of this source code is governed by the PolyForm Shield 1.0.0 license
+ * that can be found in the licenses directory at the root of this repository, also available at
+ * https://polyformproject.org/wp-content/uploads/2020/06/PolyForm-Shield-1.0.0.txt.
+ */
+
+.runtimeList {
+  padding-left: var(--spacing-5);
```
crystal-fong, line 9: wonder should all the `--spacing-number` be documented in Storybook

src/modules/75-cf/strings/strings.en.yaml
```diff
@@ -507,6 +507,9 @@ pipeline:
     toTargetGroup: to target group
     toTargetGroups: to target groups
     pleaseSelectAFeatureFlag: Please select an environment and feature flag.
+    noFlagChanges: Please add a flag change.
+    removeFlagChange: Remove flag change
+    addFlagChange: Add Flag Change
```
crystal-fong, line 512: Should this match removeFlagChange?
```suggestion
    addFlagChange: Add flag change
```

src/modules/75-cf/components/PipelineSteps/FlagConfigurationStep/FlagChangesV2/SubSection.module.scss
```diff
@@ -0,0 +1,21 @@
+/*
+ * Copyright 2023 Harness Inc. All rights reserved.
+ * Use of this source code is governed by the PolyForm Shield 1.0.0 license
+ * that can be found in the licenses directory at the root of this repository, also available at
+ * https://polyformproject.org/wp-content/uploads/2020/06/PolyForm-Shield-1.0.0.txt.
+ */
+
+.subSection {
+  border: 1px solid var(--grey-200);
+  border-radius: 5px;
```
crystal-fong, line 10: does it make a big difference if it was 4px so we can use `var(--spacing-xsmall)`?

src/modules/75-cf/components/PipelineSteps/FlagConfigurationStep/FlagChangesV2/SubSection.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+import React, { FC, MouseEvent, PropsWithChildren } from 'react'
+import { Button, ButtonSize, ButtonVariation, Container, Heading, Layout } from '@harness/uicore'
+import { FontVariation } from '@harness/design-system'
+import { useStrings } from 'framework/strings'
+import css from './SubSection.module.scss'
+
+export interface SubSectionProps extends PropsWithChildren<unknown> {
+  title: string
+  onRemove?: () => void
```
crystal-fong, line 16: if this receives onRemove from SubSections would this still be optional if it's not in SubSections?

src/modules/75-cf/components/PipelineSteps/FlagConfigurationStep/FlagChangesV2/__tests__/FlagChangesForm.test.tsx
```diff
@@ -0,0 +1,117 @@
...
+          prefix={fieldName => fieldName}
+          initialInstructions={get(initialValues, 'spec.instructions')}
+          selectedFeature={mockFeature}
+          {...props}
+        />
+      </Formik>
+    </TestWrapper>
+  )
+
+const totalSubSections = allSubSections.length
```
crystal-fong, line 55:
```suggestion
const totalSubSections = allSubSections.length

```

src/modules/75-cf/strings/strings.en.yaml
knagurski: This capitalisation matches the mocks, so that's why I went that way. Honestly, I don't flip'n know what's right these days lol

src/modules/75-cf/components/PipelineSteps/FlagConfigurationStep/FlagChangesV2/SubSection.tsx
knagurski: In an upcoming PR this will be more obvious, but when sub sections render in a Runtime context, they won't be removable. So it'll make more sense then 👍🏻

src/modules/75-cf/components/PipelineSteps/FlagConfigurationStep/FlagChangesV2/__tests__/FlagChangesForm.test.tsx
knagurski: lol, this is the kinda thing I'd highlight too

src/modules/75-cf/components/PipelineSteps/FlagConfigurationStep/FlagChangesV2/SubSection.module.scss
knagurski: What, and lose an entire pixel?!?! Are you mad?

src/modules/75-cf/components/PipelineSteps/FlagConfigurationStep/FlagChangesV2/FlagChanges.module.scss
knagurski: Difficult one. On one hand they are available to devs so they should know about them. On the other, they don't come from "Harness", but rather Blueprint (I think), so they shouldn't be suggested but only reached for in certain circumstances. This is a copy-paste of code I wrote nearly 2 years back, so I'm not sure if there's a more suitable Harness version.

src/modules/75-cf/strings/strings.en.yaml
crystal-fong: sure I know we don't have consistency across the module but the least we can do is be consistent within a section/a page?

knagurski: I'll change it coz I agree with you, but I suppose the difference is that the other uses are in a bit of info text and the other in an unseen aria attribute and then this one is used in a call to action button, so their intent is different.

src/modules/75-cf/components/PipelineSteps/FlagConfigurationStep/FlagChangesV2/FlagChanges.module.scss
```diff
@@ -0,0 +1,21 @@
...
+  margin-bottom: 0;
+
+  * + * {
```
bgribben-harness, line 12: Are you reviving the lost art of text emojis? 

But for real, is this selecting all sibling elements bar the first?

knagurski: Indeed, it's known as the [lobotomised owl](https://alistapart.com/article/axiomatic-css-and-lobotomized-owls/) pattern. It's selects all elements that have a previous sibling.

src/modules/75-cf/components/PipelineSteps/FlagConfigurationStep/FlagChangesV2/FlagChanges.tsx
```diff
@@ -0,0 +1,56 @@
...
+export interface FlagChangesProps {
+  selectedFeature?: Feature
+  initialInstructions?: FeatureFlagConfigurationInstruction[] | typeof RUNTIME_INPUT_VALUE
+  pathPrefix?: string
+}
+
+const FlagChanges: FC<FlagChangesProps> = ({ selectedFeature, initialInstructions, pathPrefix = '' }) => {
+  const { values } = useFormikContext<FlagConfigurationStepFormDataValues>()
+  const { getString } = useStrings()
+  const prefix = useCallback<(path: string) => string>(
```
bgribben-harness, line 28: Following this value through the components and their props was a bit confusing for my smooth brain because it switches from being the verb to the noun (for the function vs the value produced) - worthwhile renaming this to `getPrefix` to make it a bit clearer?

bgribben-harness: Also boo to you and your useCallbacks

knagurski: Without `useCallback` this breaks utterly. `useCallback` stabilises references and makes things safe in dependency array, which this is used in all over the show. React _should_ do this for you, but it's dumb and `useCallback` _will_ save you from CFDs

src/modules/75-cf/components/PipelineSteps/FlagConfigurationStep/FlagChangesV2/FlagChangesForm.tsx
```diff
@@ -0,0 +1,118 @@
...
+  const availableSubSections = useMemo<SubSectionComponent[]>(
+    () => allSubSections.filter(section => !subSections.includes(section)),
+    [subSections]
+  )
+
+  const onAddSubSection = useCallback((subSection: SubSectionComponent): void => {
+    setSubSections(currentSubSections => [...currentSubSections, subSection])
+  }, [])
+
+  const removeSubSection = (subSection: SubSectionComponent): void => {
```
bgribben-harness, line 77: It makes me sad that above there's an `*on*Add...` and but this is just `remove...`

src/modules/75-cf/components/PipelineSteps/FlagConfigurationStep/FlagChangesV2/FlagChanges.tsx
knagurski: Also, I totally agree about the naming, but it doesn't give you a prefix, it prefixes what you give it. I'm leaning towards `withPrefix`, but I always associate that naming with HOCs and routing, so some part of my brain doesn't like it.

src/modules/75-cf/components/PipelineSteps/FlagConfigurationStep/FlagChangesV2/FlagChangesForm.tsx
knagurski: Fair point 🤔

src/modules/75-cf/components/PipelineSteps/FlagConfigurationStep/FlagChangesV2/FlagChanges.module.scss
Harness-JKhattak: There's `--spacing-medium` and others like that which comes from uicore

https://github.com/harness/uicore/blob/main/packages/uicore/src/styles/sizes.css

src/modules/75-cf/components/PipelineSteps/FlagConfigurationStep/FlagChangesV2/SubSections.tsx
```diff
@@ -0,0 +1,44 @@
...
+    <>
+      {!subSections.length && (
+        <Container className={subSectionCSS.subSection} padding="large">
+          {getString('cf.pipeline.flagConfiguration.noFlagChanges')}
+        </Container>
+      )}
+
+      {subSections.map((SubSection, index) => (
+        <SubSection
+          key={SubSection.name}
```
bgribben-harness, line 34: Will there only ever be a max of one subsection per type in this array i.e. will this `name` value always be unique?

knagurski: Yup. Max one of each type can be added

src/modules/75-cf/components/PipelineSteps/FlagConfigurationStep/FlagChangesV2/subSections/SetFlagSwitch.tsx
```diff
@@ -0,0 +1,56 @@
...
+  })
+
+const SetFlagSwitch: SubSectionComponent = ({ prefix, ...props }) => {
+  const { getString } = useStrings()
+  const { setFieldValue } = useFormikContext()
+
+  useEffect(() => {
+    setFieldValue(prefix('identifier'), `${CFPipelineInstructionType.SET_FEATURE_FLAG_STATE}Identifier`)
+    setFieldValue(prefix('type'), CFPipelineInstructionType.SET_FEATURE_FLAG_STATE)
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
bgribben-harness, line 33: bruh

src/modules/75-cf/components/PipelineSteps/FlagConfigurationStep/FlagChangesV2/SubSection.module.scss
```diff
@@ -0,0 +1,21 @@
...
+  background-color: var(--white) !important;
+  margin-top: var(--spacing-medium);
+
+  //& > * > * {
+  //  max-width: 486px;
+  //}
```
Harness-JKhattak, line 16: Commented out CSS here

src/modules/75-cf/components/PipelineSteps/FlagConfigurationStep/FlagConfigurationInputSetStep.tsx
```diff
@@ -96,6 +104,18 @@ const FlagConfigurationInputSetStep = connect<FlagConfigurationInputSetStepProps
...
+    const hasFlagChangesRuntimeInputs = useMemo<boolean>(() => {
+      if (!Array.isArray(template?.spec?.instructions)) {
+        return false
+      }
+
+      return !!template?.spec.instructions.some(instruction => {
+        return instruction.spec?.state === RUNTIME_INPUT_VALUE // set flag state
+      })
+      // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, [flattenedInstructions])
```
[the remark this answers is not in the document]
knagurski: So React does shallow reference checks on things in a dependency array. That means that if the contents of arrays or objects change, then React is totally blind to the change. If I were to use the referenced values in the array and not the flatted form, React would totally fail to react. By flattening to a string, then react understands when the value changes and will do the right thing.

src/modules/75-cf/components/PipelineSteps/FlagConfigurationStep/FlagChangesV2/SubSection.module.scss
knagurski: Yeah, this is a hold over from v1 of the file. I was considering the impact of removing it, but don't have enough sub sections yet to prove one way or the other. Before v2 gets out into the wild, I'll have reinstated it, or killed it off. Thanks for putting it back on my radar though

src/modules/75-cf/components/PipelineSteps/FlagConfigurationStep/FlagChangesV2/FlagChanges.tsx
knagurski: That's actually the point of it. If I used context, it wouldn't be correct as it would be static. I need it to change as it goes down the layers to pick up different attributes along the way; otherwise I'd need to pass along a bunch of other props. And I can't hard code because the shape changes depending on where the step is in the pipeline and whether it's being rendered in the edit view, runtime view or input set view.

bgribben-harness: Ah I missed its usage in the effect down the page (although it has its deps array rule disabled so not sure it would make a difference).

The naming problem has an obvious solution - `prefixPrefix`

knagurski: And then when that prefix gets prefixed, I can call it `prefixedPrefixPrefix`!

src/modules/75-cf/components/PipelineSteps/FlagConfigurationStep/FlagChangesV2/subSections/SetFlagSwitch.tsx
bgribben-harness: Are there no checks you can add within the effect to ensure it only sets the values when it should? Disabling these rules is a massive code smell to me - I might spin this branch up and see this with my own eyes to believe this is the only solution lol.

knagurski: This is a super-common practice and recommended by the React team when you want something to run precisely once.

src/modules/75-cf/components/PipelineSteps/FlagConfigurationStep/FlagChangesV2/FlagChanges.module.scss
laura-mannus: wow

src/modules/75-cf/components/PipelineSteps/FlagConfigurationStep/FlagChangesV2/__tests__/FlagChangesForm.test.tsx
laura-mannus: am I blind? what is wrong here lol

laura-mannus (approved):

src/modules/75-cf/components/PipelineSteps/FlagConfigurationStep/FlagChangesV2/__tests__/FlagChangesForm.test.tsx
knagurski: @laura-mannus just a good place for a line break.

Harness-JKhattak (approved):

PR #17781, opened 2023-09-07
feat: [CDS-77667]: Add hook to log network events

rameshx:
This adds a custom hook to log network events that will be useful to debug this bugsnag issue: https://harness.atlassian.net/browse/CDS-77667


src/framework/app/bootstrap.tsx
```diff
@@ -41,6 +41,10 @@ export default async function render(): Promise<void> {
           return false
         }
 
+        event.addMetadata('NETWORK STATUS', {
+          online: navigator.onLine
+        })
```
abhinavrastogi-harness, line 46:
```suggestion
        event.addMetadata('networkStatus', navigator.onLine ? 'online' : 'offline')
```

abhinavrastogi-harness: wont something like this be better?

rameshx: The string `'NETWORK STATUS'` will be added as a tab in the bugsnag UI  so camelcase will not look good. 

The second argument needs to be an object so that it will be rendered like below:
image

abhinavrastogi-harness (approved):

mayankVermaHarness (approved):

PR #17861, opened 2023-09-12
fix: [SMP-2015]: add redirect to next-gen-ui if cg is not enabled

arya-harness:
#### Screenshots

NA

#### PR Checklist

<details>

</details>

<details>

- Jest: `retrigger jest`
- ESLint: `retrigger eslint`
- Sonar: `retrigger sonar`
- Build: `retrigger build`
- Rebase: `trigger rebase`
</details>


shaswatdeep (approved):

arya-harness:
retrigger titlecheck

chart/values.yaml
```diff
@@ -5,6 +5,8 @@ global:
   kubeVersion: ""
   loadbalancerURL: ""
   imagePullSecrets: []
+  cg:
+    enabled: false
```
abhinavrastogi-harness, line 9: why is the default value false? what happens to existing setups when this chart gets picked up?
even internally we usually have cgui enabled.

chart/templates/ingress.yaml
```diff
@@ -32,4 +32,38 @@ spec:
...
+    - host: {{ . | quote }}
+      http:
+        paths:
+          - backend:
+              service:
+                name: next-gen-ui
+                port:
+                  number: 80
+            path: /
+            pathType: Prefix
```
abhinavrastogi-harness, line 58: this doesn't looks like a redirect. 
you are just loading ngui on `/`. This causes many unexpected behaviours.
NGUI expects to, and has been tested to, run on `/ng` path only.

chart/values.yaml
arya-harness: The default value is false because in most onpremise situations, the cg is not deployed.
Since, helm charts are in umbrellas based fashion, global overrides like global.cg.enabled can be changed by customer when they want to enable it.

chart/templates/ingress.yaml
arya-harness: Yes, redirect might not be the correct word.
However, without this change, if we just hit the loadbalancer it redirects to default backend.
Since ingresses follow the rule of longest prefix match, it gets routed to respective services. If it does not find any other route it will redirect to ng-ui which inturn redirects it to gateway.
This setting used to work in onprem install for past few releases: https://github.com/harness/helm-charts/blob/release/0.8.0/src/harness/templates/ingress.yaml#L798-L829.

abhinavrastogi-harness: just to be clear, a `redirect` would mean the user also sees the new url. what you're doing here is proxying the request on the same url.
we should not do that for ngui on `/`.
NGUI should not get requests for `/` path. 
we should issue a http 302 redirect from `/` to `/ng` so that the browser makes a new request. then `/ng` will get fowarded/proxied to NGUI service correctly.

arya-harness: Adding comments from @kapilgarg1996 
we can but in this release we are keeping manifests same as last SMP release to prevent any regressions for SMP customers (given that we migrated helm charts to service repos). Moving forward, we can work on more effective solutions
---
I created a ticket in backlog to understand the differences and how we can resolve them

We tested this locally, Merging these changes as this are required for upcoming release where customer can install only ng not cg

PR #18020, opened 2023-09-19
fix: [IAC-1081]: add all path params for rendering iacm mfe in nav2.0

bgribben-harness:
- Adds additional path params for rendering IACM MFE

This fixes an issue when rendering the `MultiTypeSecretInput` from core-ui in the IACM MFE.

**Before**
image

**After**
image


Harness-JKhattak (approved):

Sagar-Rajput7 (approved):

src/modules/10-common/RouteDefinitionsV2.ts
```diff
@@ -967,6 +967,7 @@ const routes = {
   toCFConfigurePath: withModeModuleAndScopePrefix<ProjectPathProps>(() => '/configurePath'),
 
   // iacm routes
+  toIACM: withModeModuleAndScopePrefix<ProjectPathProps>(() => '/'),
```
abhinavrastogi-harness, line 970: why is `toIACM` pointing to `/`? it should be `/iacm` at the minimum.

bgribben-harness: I was advised by @Sagar-Rajput7 that for v2 navigation we did not need module prefixes in route definitions as that was taken care of by `withModeModuleAndScopePrefix`.

If I change this to be `/iacm` the module no longer renders under the expected routes.

avinashmadhwani02: @bgribben-harness You can use toModule route or toMode and pass the module name. It will be same as `/iacm`. 
Please check how it is done in CD

abhinavrastogi-harness: As discussed, we can take this up separately

abhinavrastogi-harness (approved):

abhinavrastogi-harness:
retrigger titlecheck

PR #18145, opened 2023-09-26
feat: [CDS-77846]: Add support for input dependencies, and add jenkins inputs

rameshx:
Add support for input dependencies, and add jenkins build step inputs

**Screenshot:**
image


src/modules/70-pipeline/y1/components/InputFactory/InputComponentType.ts
```diff
@@ -16,8 +16,10 @@ export enum PrimitiveInputType {
   timeout = 'timeout'
 }
 
-export enum DerivedComponentType {
-  ShellScript = 'shell_script'
+export enum DerivedInputType {
+  shell_script = 'shell_script',
+  jenkins_job_name = 'jenkins_job_name',
+  jenkins_connector = 'jenkins_connector'
```
srdjan-harness, line 22: Do we need jenkins connector?  it should be connector with subtype.

srdjan-harness (approved):

src/modules/70-pipeline/y1/components/InputFactory/InputComponentType.ts
rameshx: We decided to use `jenkins_connector` for now.

PR #18223, opened 2023-09-29
fix: [CDS-79421]: Update FormMultiTypeConnectorField to support any CSS width

rameshx:
FormMultiTypeConnectorField only supports fixed width values, making it harder to use the component where relative widths are needed. This PR updates it to support any CSS width.

| Before | After |
|--------|--------|
| image | image |


src/modules/27-platform/connectors/components/ConnectorReferenceField/FormMultiTypeConnectorField.tsx
rameshx, line 1: Ideally, the default width should be 100%, but not updating it from 400 (line 123) to avoid breaking other usages.

PratyushGar (approved):

vikrant-harness (approved):

pankajsingh-harness (approved):

d1wilko (approved):
LGTM

mayankVermaHarness (approved):

PR #18889, opened 2023-11-09
techdebt: [PL-39227]: do not allow internal snapshots of more than 25 lines and do not allow external snapshots at all

chetanvekariya-harness:
### Summary
**Problem:**
- We have discussed many times that we need to reduce snapshot testing because snapshot testing is heavily exploited by us to get coverage check pass and then they are there to give us false confidence that we have covered entire feature under jest test cases while in reality almost no functionality is covered. Additionally, when somebody makes changes and if snapshots break we tend to update them blindly without debugging the root cause. Several times, the change is actually an issue and should have been fixed in code, but we end up updating snapshots which means we have covered a bug under snapshot test and we expect it to be present. Because of all these issues, we want to reduce snapshot testing.

**Solution:**
- jest/no-large-snapshots eslint rule allows us to provide internal and external snapshot size. This works for internal snapshots, but does not work for external, hence wrote pre-commit husky rule that prevent usage of `toMatchSnapshot`.
- From now on, if snapshot testing is required in rare cases then inline snapshot of less than 25 lines will be allowed. No external snapshot allowed at all.

**Ref for eslint rule** - https://github.com/jest-community/eslint-plugin-jest/blob/main/docs/rules/no-large-snapshots.md

**Articles explaining why snapshot testing should be used sparsely:**

https://medium.com/@sapegin/whats-wrong-with-snapshot-tests-37fbe20dfe8e#:~:text=Test%20intentions%20are%20hard%20to,the%20behavior%20remains%20the%20same.

https://kentcdodds.com/blog/effective-snapshot-testing

#### Screenshots
NA

#### PR Checklist

<details>

</details>

<details>

- Jest: `retrigger jest`
- ESLint: `retrigger eslint`
- Sonar: `retrigger sonar`
- Build: `retrigger build`
- Rebase: `trigger rebase`
- MessageMetadata: `trigger messagecheck`
- GitLeaks: `trigger gitLeaks`
</details>

## PR check failures and solutions
https://harness.atlassian.net/wiki/spaces/BT/pages/21106884744/PR+Checks+-+Failures+and+Solutions

### Rebase Instructions:
To rebase your branch onto the latest changes in the target branch, simply use the following command in a comment: `/rebase`


chetanvekariya-harness:
retrigger eslint

chetanvekariya-harness:
trigger messagecheck

chetanvekariya-harness:
trigger messagecheck

src/modules/70-pipeline/components/RunPipelineModal/__tests__/useInputSets.test.tsx
```diff
@@ -184,67 +184,65 @@ describe('<useInputSets /> tests', () => {
...
+                          name: 'var1',
+                          type: 'String',
+                          value: '123'
+                        }
+                      ]
                     },
-                  },
-                },
-                "type": "Deployment",
+                    type: 'Kubernetes'
```
abhinavrastogi-harness, line 206: is this change intentional? doesn't seem related to any code changes

chetanvekariya-harness: There is no real change here. Usage of `toMatchInlineSnapshot` has been removed with `toEqual`.

.husky/no-external-snapshots
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
+
+gitDiff="$(git diff --staged --diff-filter=M -- '*.test.*' --unified=0 | egrep "^(\+) " | grep -w '.toMatchSnapshot\|.toThrowErrorMatchingSnapshot')"
+
+if [[ -n "$gitDiff" ]]; then
```
vikrant-harness, line 5: this will give error in case when in existing test `toMatchSnapshot` is present and the line number changes ?

chetanvekariya-harness: I will check it myself, but Is there a way to avoid it? This also happens for our other pre-commit hook that checks for formName and data-tooltip-id change.

vikrant-harness: yes it happens for that as well , but this will be blocking PR's frequently so people will use -f , we need to catch it in PR's then

chetanvekariya-harness: You are right. We need to catch this PR for which I am going to implement eslint rule, but that can be done once we do not have any external snapshot. This will block commit as frequently as other pre-commit hook, I guess. There is no issue doing `-f`, if you think you have not made mistake. Moreover, we are supposed to prevent snapshot testing while reviewing PRs. We should block PRs if somebody does snapshot testing.

abhinavrastogi-harness: can you check if the condition `--diff-filter=A` will give only newly added files? that might be a safer approach to allow snapshots in existing files, but disallow in new files.

chetanvekariya-harness: Sure, Abhinav. I can check that. But, we would want to disallow any new addition of external snapshot irrespective of new/old file. So, I think we need to check all new changes, not just new files' changes.

abhinavrastogi-harness: In that case, we disallow addition or modification of `.snap` files instead of checking for `.toMatchSnapshot`. What do you think?

chetanvekariya-harness: Initially, I was trying for that and faced an issue. But, the issue was for eslint rule. It should be easy to do with pre-commit hook. Let me try this. Thanks for suggestion.

chetanvekariya-harness: @abhinavrastogi-harness - Made changes as per your comment. Tested in local.
This will give error for all additions. Only time this will be irritating is when we update existing snapshot, to handle it, one has to either force commit or remove snapshot and write assertions. 
But, this is really helpful when someone adds to existing snapshot, they will get error for adding new content to existing snapshot because we want to prevent that as well.
But, if you still think we should run it only for newly added snapshot files. Let me know, I can make the change.

knagurski: Regrettably, we can't really do it on all snapshot updates. Otherwise when we next update anything in a base component (e.g. add a `data-testid` in a component from UI Core), then we'll need to purge most of the snapshots of NGUI... that sounds great to me, but it's not practical (regrettably). Instead of new files, can we look at the specific blocks in the snapshot files and prevent new blocks? Otherwise we won't catch new snapshots in existing test suites.

abhinavrastogi-harness (approved):

.husky/no-external-snapshots
chetanvekariya-harness: @knagurski - I think using `--diff-filter=AM` picks up all the snapshots and then we are looking for all the lines with `+` which picks up all snapshots changes. As I mentioned, if change is required for the cases like you mentioned above, we can go ahead and force-commit. 
Else, we will have to use `--diff-filter=A` which will allow all kinds of modifications and will block only for new additions. Should we do this task just for new additions?
I am not sure if it is even possible to catch just new additions and additions to existing snapshots. Do you think it is possible? I will still explore bit more and see if we can allow modifications to existing snapshots.

knagurski: It's a difficult one. I kinda like the brute force all-or-nothing approach, but my concern is that we might shift normalising hundred-plus file blind updates, to that _PLUS_ an over use of force-merge. At least without this, people who trigger huge snapshot updates will still have to seek out code owners and can eventually merge normally; when we pretty much require a force-merge people might skip that part and push for a force-merge when sign-offs aren't complete or other things are failing.

I don't have a good answer unfortunately, beyond parsing the snapshot files manually and determining if the new snapshot name exists already... if it exists, then we (grudgingly) let it past, otherwise we can throw the fail. But that's a lot more work. I'd say that we're probably best going with what you have and just being mindful of the potential issues and ask those with force-merge powers to be vigilant.

chetanvekariya-harness: @knagurski - I think there is bit of misunderstanding here. We do not need force-merge at all. We have written pre-commit hook here, so it will be just force-commit but that should only be done for existing snapshots updates. We will request everyone not to do force-commit for new snapshots additions / for addition to existing snapshot files and if someone does it, reviewers should catch it and block the PR, because snapshots any way will have to be migrated in near future once we come up with authors list who will be responsible for deleting snapshots and write assertion based testing, so it does not make sense to add new snapshots which will be required to be deleted in near future.

vikrant-harness (approved):

.husky/no-external-snapshots
knagurski: Sorry, you're 100% correct. The point does still stand that people might get in the habit of bypassing the pre-commit checks (its so much easier to add `--no-verify` than to wait for a failure). But the PR checks will catch most things after that 👍🏻

PR #18970, opened 2023-11-15
feat: [CDS-83891]: Saving Grid or List view type to the preference store

avinashmadhwani02:
#### Screenshots

NA

#### PR Checklist

<details>

</details>

<details>

- Jest: `retrigger jest`
- ESLint: `retrigger eslint`
- Sonar: `retrigger sonar`
- Build: `retrigger build`
- Rebase: `trigger rebase`
- MessageMetadata: `trigger messagecheck`
- GitLeaks: `trigger gitLeaks`
</details>

## PR check failures and solutions
https://harness.atlassian.net/wiki/spaces/BT/pages/21106884744/PR+Checks+-+Failures+and+Solutions

### Rebase Instructions:
To rebase your branch onto the latest changes in the target branch, simply use the following command in a comment: `/rebase`


avinashmadhwani02:
trigger messagecheck

src/modules/72-templates-library/pages/TemplatesPage/TemplatesPage.tsx
```diff
@@ -74,15 +75,20 @@ import RepoFilter from '@common/components/RepoFilter/RepoFilter'
...
+import { PreferenceScope, usePreferenceStore } from 'framework/PreferenceStore/PreferenceStoreContext'
 
 import css from './TemplatesPage.module.scss'
 
 export default function TemplatesPage(): React.ReactElement {
   const { getString } = useStrings()
   const history = useHistory()
   const queryParamOptions = useTemplatesQueryParamOptions()
-  const { templateType, repoName, page, view, size, sort, searchTerm, filterIdentifier, filters } =
+  const { templateType, repoName, page, size, sort, searchTerm, filterIdentifier, filters } =
```
pankajsingh-harness, line 86: We should still consider `view` type from the URL and also from the user preference. 
Preference should be given to `view` type from the URL.

pankajsingh-harness: There may be some URLs bookmarked having a `view` type.

avinashmadhwani02: I believe it's better to maintain a single source. If we store it in local storage, there's no need for a view type in the URL

For bookmark urls to work - we will have to add one useEffect and set it in localStorage. I think we can avoid that, since the source is going to be local storage in the future.

what are your thoughts?

pankajsingh-harness (approved):

src/modules/72-templates-library/pages/TemplatesPage/TemplatesPage.tsx
mayankVermaHarness: I have seen elsewhere on Harness (CC: @satyam-harness can confirm) that `view` (or other query params) from URL were given preference over the one saved in PreferenceStore (local storage).

satyam-harness: Yes, view in url is given more preference, its in the log view of execution page. If some user has view1 stored in localStorage and has a url which has view1 in it, and opens the url then view1 should appear.

avinashmadhwani02: checked with @satyam-harness , the flow is not consistent on Executions page also since there are different sources (we don't save the view to preference store if the User has visited the page directly through the url)

To avoid the different sources, I was thinking it's better to keep local storage as only source. (only bookmark url use case will not work here - looks like a very corner case).

mayankVermaHarness (approved):

PR #19022, opened 2023-11-20
fix: [PL-21367]: Tags are not displayed on mouse hover on Filter icon

ishantmehtaharness:
description about the change:
tags ia an object here, and the filter component needs it as an array to be render it

required screen shots or Videos about the fix
Attached screenshot

area impacted with the fix
This change is specific to Filter in connectors, so no other area should be impacted

test cases tested for the fix
I have tested it for empty tags, and another scenario with tags.
API calls are also working fine in both the cases

Screenshot 2023-11-20 at 2 13 15 PM


ishantmehtaharness:
trigger messagecheck

ishantmehtaharness:
trigger messagecheck

ishantmehtaharness:
trigger messagecheck

ishantmehtaharness:
trigger messagecheck

litvinovstas (approved):

pankajsingh-harness (approved):

namanharness (approved):

PR #19187, opened 2023-11-29
feat: [FFM-9960]: Allow editing of flag changes v1 pipelines

knagurski:
This PR adds backwards compatibility with pipelines created with v1.


knagurski:
trigger messagecheck

crystal-fong (approved):

src/modules/75-cf/strings/strings.en.yaml
```diff
@@ -505,6 +505,16 @@ pipeline:
...
     flagChanges: Flag Changes
+    flagChangesV2Runtime: >
+      <p>At runtime, you may choose the following flag configuration changes:</p>
+      <ul>
+        <li>Set flag switch</li>
+        <li>Set default ON rule</li>
+        <li>Set default OFF rule</li>
+        <li>Serve variation to targets</li>
+        <li>Serve variation to target groups</li>
+        <li>Serve percentage rollout to target group</li>
```
laura-mannus, line 516:
```suggestion
        <li>Serve percentage rollout to target groups</li>
```
should this be plural?

laura-mannus (approved):

src/modules/75-cf/strings/strings.en.yaml
knagurski: Nope, you can only set for a single target group. Funnily enough, I had it as plural on accident and only spotted the error just before I committed.

PR #19367, opened 2023-12-12
feat: [FFM-10189]: Re-introduce bucket by

knagurski:
This PR re-introduces the Bucket By field to places that display the percentage rollout component.


Kira-Jennings (dismissed):

laura-mannus (approved):

laura-mannus (approved):

src/modules/75-cf/components/PipelineSteps/FlagConfigurationStep/FlagChangesV2/subSections/ServePercentageRolloutToTargetGroup/__tests__/BucketByField.test.tsx
```diff
@@ -0,0 +1,103 @@
...
+    expect(await screen.findByPlaceholderText('cf.percentageRollout.bucketBy.placeholder')).toBeInTheDocument()
+
+    await userEvent.click(await screen.findByRole('button'))
+    await userEvent.click(await screen.findByText('Runtime input'))
+
+    expect(screen.queryByPlaceholderText('cf.percentageRollout.bucketBy.placeholder')).not.toBeInTheDocument()
+    expect(await screen.findByPlaceholderText(RUNTIME_INPUT_VALUE)).toBeInTheDocument()
+  })
+
+  test('it should fetch target attributes when the env, project, org and account IDs are available', async () => {
```
crystal-fong, line 59: don't have to address right now but could we also assert that if either env/project/org/account is missing, then it will default target attributes to be an empty array?

knagurski: I had thought about that. I didn't add a test here because the drop down won't be empty, it'll have the 2 defaults. I suppose I could assert that the hook was called with `lazy: true` to prove it didn't try to do the call 🤔 Either that or look at the menu itself and could the options in it

src/modules/75-cf/strings/strings.en.yaml
```diff
@@ -2227,5 +2228,10 @@ cfTrialHomePage:
     startBtn:
       description: Start 14 day FF Enterprise trial
 percentageRollout:
-  toTargetGroup: 'To Target Group:'
+  toTargetGroup: 'To target group'
```
crystal-fong, line 2231: nbd but these quotes aren't needed

crystal-fong (approved):

src/modules/75-cf/components/PipelineSteps/FlagConfigurationStep/FlagChangesV2/subSections/ServePercentageRolloutToTargetGroup/__tests__/BucketByField.test.tsx
crystal-fong: ye, pretty cool lazy trick as well

src/modules/75-cf/strings/strings.en.yaml
knagurski: True. Needed with the `:` in there, but not without

PR #19442, opened 2023-12-14
fix: [CDS-86829]: Fix Styling - file path should be wrapped. its goin…

ishantmehtaharness:
Before

Before

After

Screenshot 2023-12-14 at 2 44 01 PM

Screenshot 2023-12-14 at 2 44 04 PM

Tests
- tested with small test
- tested with big text - it shows truncated and shows tooltip


src/modules/70-pipeline/components/ManifestSelection/__tests__/__snapshots__/ManifestSelection.test.tsx.snap
```diff
@@ -990,6 +991,7 @@ exports[`ManifestSelection tests renders add Manifest option without crashing 1`
                       <div
                         class="StyledProps--font StyledProps--main StyledProps--flex StyledProps--flex-justifyContent-flex-start"
                         data-testid="1"
+                        style="flex: 1;"
```
vikrant-harness, line 994: @ishantmehtaharness let's remove the snapshot testing with the same PR ?

ishantmehtaharness: Removing will reduce the coverage

ishantmehtaharness: And I see this file only has snapshots. We should plan to refactor that and remove snapshots

vikrant-harness: Yes we need to remove snapshots and write proper assertions even if it reduces coverage initially, but sure you can create one ticket and tag it here Thanks! CC :- @chetanvekariya-harness

vikrant-harness (approved):

satyam-harness (approved):

